PR #240, opened 2026-05-09
fix(rds): flip StorageEncrypted default to true (CIS-AWS RDS.3)

Cre-eD:
## Summary

PR #239 added the `StorageEncrypted *bool` field as **opt-in** (nil → false → unencrypted), preserving pre-2026.5 behaviour. That left every customer who hadn't read the new field's docs on unencrypted RDS, violating **CIS-AWS Foundations RDS.3**.

This 1-line-each change flips the resolver from \`lo.FromPtr(*ptr)\` to \`lo.FromPtrOr(*ptr, true)\` at both call sites, making encryption the secure default while preserving every existing customer's data.

| State | Before | After |
|---|---|---|
| `nil` (omitted from YAML) | false (unencrypted) | **true** (encrypted) |
| `&true` (explicit) | true | true |
| `&false` (explicit) | false | false |

## Why existing DBs are safe — IgnoreChanges already in place

PR #239 added \`sdk.IgnoreChanges([]string{\"storageEncrypted\"})\` on the resource opts. Pulumi's IgnoreChanges semantics: *"treat the desired value as the recorded state value for the diff."*

| Scenario | State | New spec | Diff | Outcome |
|---|---|---|---|---|
| Existing unencrypted, never set field | `false` | `true` (default flip) | IgnoreChanges silences | **No replacement, no data loss** |
| Existing encrypted, set `true` previously | `true` | `true` | no change | unchanged |
| Customer explicitly sets `false` in YAML | * | `false` | IgnoreChanges silences | honoured, no second-guessing |
| Brand-new stack | none | `true` | initial create | encrypted ✓ |

Customers who genuinely want to *migrate* an existing unencrypted RDS to encrypted still need the out-of-band path AWS forces (snapshot → encrypted-copy → restore → re-import). The IgnoreChanges deliberately blocks SC from attempting that migration via destructive replacement.

## Tests

`pkg/clouds/aws/rds_storage_encrypted_test.go` updated:
- omit-case description: \"legacy default, encryption off\" → \"secure default, encryption on\"
- resolver: \`lo.FromPtr(cfg.StorageEncrypted)\` → \`lo.FromPtrOr(cfg.StorageEncrypted, true)\`
- expected: \`wantSet && wantVal\` → \`!wantSet || wantVal\`

Locally: \`go build ./...\`, \`go vet ./...\`, \`TestReadRds{Mysql,Postgres}Config_StorageEncrypted\` all pass.

## Companion

[simple-container-com/actions#9](https://github.com/simple-container-com/actions/pull/9) tightens the \`go-aws-rds-no-storage-encryption\` Semgrep rule to whitelist **only** the \`lo.FromPtrOr(_, true)\` shape — bare \`lo.FromPtr(*ptr)\` and \`lo.FromPtrOr(*ptr, false)\` both still fire the rule.

After both land:
- This api PR's Semgrep status passes (post-fix code uses the whitelisted shape).
- Any regression to bare \`lo.FromPtr(*bool)\` on a security-critical default gets flagged at PR time across SC + every consumer repo using the ruleset.

## Test plan

- [x] go build ./... + go vet ./... locally
- [x] TestReadRds{Mysql,Postgres}Config_StorageEncrypted green
- [ ] CI green
- [ ] (after actions#9 merges) re-scan: 0 findings of `go-aws-rds-no-storage-encryption`
- [ ] Validate via SC preview build against an existing PAY-SPACE staging stack — confirm `pulumi preview` shows no destructive replacement on the existing unencrypted RDS instances

### DIFF
--- a/pkg/clouds/aws/rds_mysql.go
+++ b/pkg/clouds/aws/rds_mysql.go
@@ -18,18 +18,21 @@ type MysqlConfig struct {
 	Password        string  `json:"password" yaml:"password"`
 	DatabaseName    *string `json:"databaseName" yaml:"databaseName"`
 	EngineName      *string `json:"engineName,omitempty" yaml:"engineName,omitempty"`
-	// StorageEncrypted opts into AWS-side encryption-at-rest for the
-	// underlying EBS volume. When unset (nil), the instance is created
-	// with the AWS default — currently UNENCRYPTED — preserving exact
-	// behaviour for stacks that pre-date this field. Set `true` to opt
-	// the resource into encryption (uses the AWS-managed `aws/rds` KMS
-	// key by default).
+	// StorageEncrypted controls AWS-side encryption-at-rest for the
+	// underlying EBS volume. When unset (nil), new instances default
+	// to ENCRYPTED (AWS-managed `aws/rds` KMS key), matching CIS-AWS
+	// Foundations RDS.3. Set `false` explicitly to opt out for legacy
+	// unencrypted stacks; set `true` to be explicit.
 	//
-	// AWS RDS `storage_encrypted` is IMMUTABLE post-creation. Toggling
-	// this field on an existing unencrypted instance does NOT migrate
-	// data — it is silenced via `pulumi.IgnoreChanges` to prevent a
-	// destructive replacement. To convert an existing unencrypted RDS
-	// to encrypted, snapshot → encrypted-copy → restore → re-import.
+	// AWS RDS `storage_encrypted` is IMMUTABLE post-creation. The
+	// default flip is safe for existing instances because the
+	// `pulumi.IgnoreChanges` on the resource opts (see
+	// pkg/clouds/pulumi/aws/rds_mysql.go) silences storage_encrypted
+	// drift — Pulumi will not propose a destructive replacement when
+	// the spec value differs from the cloud-actual value. Customers
+	// who want to genuinely migrate an existing unencrypted RDS to
+	// encrypted must do it out-of-band: snapshot → encrypted-copy →
+	// restore → re-import.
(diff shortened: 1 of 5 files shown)


Cre-eD:
## ✅ Empirical validation passed

Ran [`sc provision --preview --diff --skip-refresh`](https://github.com/Integrail/devops/actions/runs/25623996102) against an existing **unencrypted** \`aws-rds-postgres\` in \`Integrail/devops/.sc/stacks/integrail/server.yaml\` (staging environment), with SC built from this PR's branch (\`v2026.5.7-pre.7436b6b-preview.7436b6b\`).

### Plan summary

| Action | Count |
|---|---|
| create | 0 |
| update | 2 |
| **replace** | 2 |
| delete | 0 |
| **unchanged** | **463** ← includes \`aws-rds-postgres\` |

The 2 replacements + 2 updates are unrelated cloudtrail-security-helpers \`docker:index/remoteImage:RemoteImage\` rebuilds (new SC version → new helper image digest). No RDS, RDS-cluster, EC2, VPC, or related resource appears in the diff.

### What this proves

- Spec resolved to \`storageEncrypted: true\` on the new code (otherwise diff couldn't have been suppressed — would've trivially matched state).
- Cloud-actual is \`storage_encrypted: false\` on the existing instance.
- \`sdk.IgnoreChanges([]string{"storageEncrypted"})\` (added in #239) **correctly silenced the spec-state drift** — Pulumi registered the resource as "unchanged" rather than "replace".
- **No data-loss path**. Default-flip is safe to merge.

Validation PR (closed, branch deleted): [Integrail/devops#364](https://github.com/Integrail/devops/pull/364)

smecsia (approved):

PR #255, opened 2026-05-11
fix(k8s+caddy): stop namespace-rename cascade-delete + Caddy fallout (PAY-SPACE 2026-05-10, fulldiveVR 2026-05-12 outages)

Cre-eD:
Two consumer outages, same root cause: SC api #230's namespace rename triggers a Pulumi Replace that cascade-deletes the shared parent namespace on the first `pulumi up` after #230 ships. Plus the Caddy fallout from that cascade was invisible to monitoring.

## Confirmed outages

- **PAY-SPACE 2026-05-10/11**: every whitelabel under `parentEnv: production` (`support-payhey`, `support-rulex`, `support-gl-pay`, parallel wallets) cascade-deleted the shared `support-bot` / wallet namespaces. Caddy then served the welcome page on every prod host as HTTP 200, hiding the outage from monitoring until a human opened a browser.
- **fulldiveVR/wizeup-rooms-api 2026-05-12**: namespace `wize-rooms-api` hosted both the `likeclaw-us` parent and the `likeclaw-us-dev` child. A routine merge to dev triggered the child's deploy. Pulumi plan: `kubernetes:core/v1:Namespace: (replace) name "wize-rooms-api" => "wize-rooms-api-likeclaw-us-dev"`. Namespace deleted, `rooms-api.wizeup.app` returned 502 until prod was manually re-deployed. ([actions/runs/25725750825](https://github.com/fulldiveVR/wizeup-rooms-api/actions/runs/25725750825))

Anyone with `parentEnv != stackEnv` whose Pulumi state predates #230 is at risk on their next `pulumi up`. Two confirmed so far; likely more.

## Root-cause fix — `IgnoreChanges("metadata.name")` on Namespace

#230 added `RetainOnDelete(true)` expecting that to protect existing consumers through the migration. It didn't: Pulumi reads delete-time options from the **state** of the resource being deleted, not from the current program. The old Namespace resource in state predates #230 and doesn't carry the flag, so `Replace` proceeds with the k8s DELETE and cascade-kills the parent.

[efd2523](https://github.com/simple-container-com/api/pull/255/commits/efd2523) adds `sdk.IgnoreChanges([]string{"metadata.name"})` to both `corev1.NewNamespace` call sites (`simple_container.go` client stacks + `helpers.go` helm operator stacks).

Behavior:
- **Fresh deploy of a new custom child stack**: no prior state → no diff to ignore → namespace created with the per-stackEnv name. PR #230's isolation goal preserved.
- **Existing custom child stack on next `pulumi up`**: state's old `metadata.Name` vs program's new desired `metadata.Name` would normally schedule a Replace; `IgnoreChanges` suppresses that diff. No Replace, no delete, no cascade. State retains the legacy name. Service / Deployment / etc. follow that name and continue to land in the shared namespace. Migration cost: zero.
- **Existing consumer who actively wants the new isolated namespace**: opt-in by removing the legacy `Namespace` resource from Pulumi state (state-edit; k8s namespace itself stays). Next `pulumi up` registers a fresh namespace at the new name.

Established codebase pattern: `rds_postgres.go:45` and `rds_mysql.go:55` use the same shape (`IgnoreChanges([]string{"storageEncrypted"})`) for the same purpose — silence a default flip so it doesn't propose a destructive replacement on existing stacks.

## Caddy fallout fixes (also in this PR)

When the cascade-delete in the root cause path finished, every Service with `simple-container.com/caddyfile-entry` for the affected hosts disappeared. Two distinct Caddy failure modes followed:

1. **Aggregator crashloop during the Replace window** — for the brief moment the old + new Services coexisted, two `http://<domain> { ... }` site blocks ended up in `/tmp/Caddyfile` and Caddy aborted with `ambiguous site definition: http://<domain>`. Commits [2e0eeae](https://github.com/simple-container-com/api/pull/255/commits/2e0eeae) + [1abd3c1](https://github.com/simple-container-com/api/pull/255/commits/1abd3c1): dedup by site-address (first non-blank, non-comment line of the annotation, whitespace-trimmed), most-recent Service wins via `creationTimestamp` + `sort -r`, `set -eo pipefail` so a flaky kubectl can't silently produce an empty config.

2. **Default catch-all served HTTP 200 + welcome page** — after the cascade finished, requests for production hosts fell through to `http:// { file_server /etc/caddy/pages }` and got `200 OK "Default page"`. External monitoring, CDNs, uptime checks all saw healthy 200s. Commits [e5a6519](https://github.com/simple-container-com/api/pull/255/commits/e5a6519) + [d7b4d71](https://github.com/simple-container-com/api/pull/255/commits/d7b4d71) + [328e796](https://github.com/simple-container-com/api/pull/255/commits/328e796): default block now returns `503` with `Retry-After: 60`, `Cache-Control: no-store`, `Content-Type: text/html`, wrapped in an explicit `handle { ... }` so headers + body apply only to the 503 path. Removed `import hsts` from the catch-all so the 503 reaches monitoring directly instead of redirecting into a TLS handshake failure for unknown SNI.

3. **Operational hardening** — [95730bf](https://github.com/simple-container-com/api/pull/255/commits/95730bf): dropped `set -x` so annotation bodies aren't traced to cluster logs.

Dead code removal: `/etc/caddy/pages/index.html` (the "Default page" template) deleted, no longer referenced. `404/500/502.html` retained — still used by per-Service `handle_*_error` snippets.

## Review provenance

This PR has been through four rounds of parallel `codex` + `gemini` review on the Caddy half. Convergent on "mergeable" in round 4. Each fixup commit captures one round's findings; commit history is intentionally not squashed so the review trail is auditable. Comments above on the PR record the round-by-round summaries.

The namespace-root-cause commit (efd2523) is fresh — needs its own review pass before merge.

## Test plan

Unit:
- `go build ./...` clean
- `go test ./pkg/clouds/pulumi/kubernetes/... -count=1` passes

Behavioral (manual, post-merge with branch preview):
- Fresh deploy of a new custom child stack: Namespace created with `<stackName>-<stackEnv>`.
- Existing PAY-SPACE / fulldiveVR consumer deploy: Pulumi diff for Namespace should show NO Replace (the `metadata.name` diff is suppressed by `IgnoreChanges`). All other resources (Service, Deployment, …) unchanged.
- Caddy: validates clean, returns 503 with right headers for unknown Host, 200 for matched site blocks. Verified in commit message of e5a6519 with real `simplecontainer/caddy:latest`.

## Followup

Memory recorded for next time: this is the second SC migration in two days where a `metadata.name` change was assumed to be safe under `RetainOnDelete`. Future SC changes to `metadata.name` of any long-lived resource should default to `IgnoreChanges` from the start, not retrofit after an outage.


Cre-eD:
## Review pass — codex + gemini (parallel)

Ran a parallel review with both tools. Headline: **codex caught a critical regression I introduced**, gemini caught some defensive concerns. Pushed [1abd3c1](https://github.com/simple-container-com/api/pull/255/commits/1abd3c1) addressing both.

### Codex — the serious one

> `kubectl get ... | sort -r` under `set -e` without `pipefail` — a kubectl error/timeout becomes `services=""` and the init-container exits successfully with only the default Caddyfile.

This automated the exact welcome-page outage scenario from 2026-05-10. Without pipefail, any transient kubectl flake on a Caddy pod restart silently produces a config with no Service routes, the catch-all `http:// { file_server }` serves the welcome page on every domain, and monitoring sees 200 OK. Hard miss on my part.

Fix in 1abd3c1:
- `set -xeo pipefail`
- Split kubectl/sort into separate assignments to make the failure path explicit even for readers who skip the `set` line

### Gemini — defensive

Flagged the dedup key as brittle in theory: leading whitespace, comment lines, multi-host comma-separated site addresses. For SC-generated annotations this is moot (first line is deterministic `${proto}://${domain} {` or `handle_path /${prefix}*`), but indentation/comment normalization is cheap defense.

Fix in 1abd3c1:
```awk
/^[[:space:]]*$/ { next }
/^[[:space:]]*#/ { next }
{ sub(/^[[:space:]]+/, ""); sub(/[[:space:]]+$/, ""); print; exit }
```

Verified offline: `  http://example.com {` (new, indented) and `http://example.com {` (old, flush) now produce the same dedup key. Comment-led annotations correctly emit with the underlying site-block as key.

Multi-host comma-order normalization (gemini's `host1, host2 {` vs `host2, host1 {`) is intentionally not addressed — not a case SC generates, and lexically sorting hosts inside a key would require a Caddyfile parser. Defer to a real follow-up if/when the use case appears.

### Both reviewers explicitly confirm

This PR fixes **only the secondary crashloop**. It does not fix the actual prod-killer (non-retroactive `RetainOnDelete` cascade-deleting the shared parent namespace on first `pulumi up` after the migration). Two follow-up PRs queued:

1. Retroactive `RetainOnDelete` path so namespace resources that entered Pulumi state before #230 don't get cascade-deleted during the Replace
2. Caddy default-block hardening: serve a hard 503 instead of `file_server /etc/caddy/pages` when no Service block matches, so route absence is loud instead of disguised as healthy 200s

Both follow-ups are real bugs. This one's safe to merge first as the immediate symptom containment.

Cre-eD:
## Multi-round codex + gemini parallel review — converged

Four rounds, two reviewers per round. Both reach "mergeable" in round 4.

### Round 1 — codex blocked the PR, gemini was too lenient

Codex caught the critical regression I'd introduced: `kubectl ... | sort -r` under `set -e` without `pipefail` would silently produce `services=""` on any kubectl flake, automating the welcome-page-masquerade outage that the PR is trying to fix. Plus:
- `respond` defaults Content-Type to `text/plain` → browsers saw raw HTML tags
- `header Cache-Control` and `header Retry-After` leaked onto the HSTS 301 path
- `import hsts` redirected catch-all HTTP→HTTPS→TLS-handshake-fail, making the 503 unreachable behind any CDN that sets `X-Forwarded-Proto`
- Stale comment + dead `index.html`

Gemini noticed the HSTS issue but called it "still loud enough." Codex was correct — TLS-fail is invisible to HTTP monitoring, which is the whole point of this fix.

**Pushed:** [d7b4d71](https://github.com/simple-container-com/api/pull/255/commits/d7b4d71) + [328e796](https://github.com/simple-container-com/api/pull/255/commits/328e796) — `set -eo pipefail`, explicit `Content-Type`, `handle { ... }` wrapper, removed `import hsts` from catch-all, deleted index.html.

### Round 2 — both clean

Codex: "Findings: none blocking. Mergeable now." Verified all five round-1 fixes with real Caddy testing across h2, prefix-mode, and per-Service-with-hsts.
Gemini: "Mergeable as-is."

One nit: pipefail comment said "kubectl piped into sort" but code now splits them. **Pushed:** [2d290fe](https://github.com/simple-container-com/api/pull/255/commits/2d290fe) — refreshed comment.

I then accidentally put markdown backticks inside the Go raw string and broke the build. **Pushed:** [9828eb5](https://github.com/simple-container-com/api/pull/255/commits/9828eb5) — fix.

### Round 3 — codex clean, gemini blocked

Gemini flagged `set -x` as a security regression: tracing every command dumps annotation bodies into cluster logs, and a consumer-misuse path (basicauth credentials in `Headers` map, raw directives in `LbConfig.ExtraHelpers`) could template secrets into the annotation. Codex noted the same exposure existed via `cat /tmp/Caddyfile` and called it pre-existing.

Sided with gemini. **Pushed:** [95730bf](https://github.com/simple-container-com/api/pull/255/commits/95730bf) — dropped `-x`, kept `set -eo pipefail`. Kept `cat /tmp/Caddyfile` since that's intentional final-config logging, useful for rollout verification, and the trace-amplification (every kubectl invocation × every Service × every pod restart) was the bigger leak.

### Round 4 — converged

Both reviewers: mergeable, no blockers. Codex confirmed `set -eo pipefail` preserves errexit + pipeline-failure behavior; only execution tracing was removed. Gemini explicitly endorsed keeping `cat /tmp/Caddyfile` as "a pragmatic trade-off" — agrees with my reasoning.

### Net result

7 commits. From original two-line dedup PR to a hardened, security-reviewed change that addresses both halves of the 2026-05-10 outage:
- aggregator dedup + pipefail + key normalization (prevents the Caddy crashloop during Service Replace)
- default catch-all returns 503 with text/html + Retry-After + Cache-Control: no-store, behind a `handle { ... }` scope that doesn't leak headers onto unrelated paths
- removed catch-all `import hsts` so the 503 reaches monitoring directly instead of redirecting into a TLS failure
- dropped `set -x` to reduce annotation-body exposure to cluster logs

Ready to merge.

Cre-eD:
## Namespace work — three review rounds (codex + gemini parallel each round)

The IgnoreChanges fix in efd2523 was incomplete. Multi-round review caught it and the follow-up is in [1d248d9](https://github.com/simple-container-com/api/pull/255/commits/1d248d9).

### Round 1 — clean, both approved

Both confirmed semantics: `IgnoreChanges("metadata.name")` suppresses the Replace for existing state, has no effect on fresh deploys, downstream resources follow `namespace.Metadata.Name().Elem()` correctly. Codex even built a throwaway Pulumi stack and tested it live.

### Round 2 — codex caught a critical bug, gemini missed it

Codex found that two callsites in simple_container.go still used the program-computed `sanitizedNamespace` (string) instead of the live `namespace.Metadata.Name()` Output:

- **caddyfile-entry annotation (line 673)** templated `sanitizedNamespace` into `reverse_proxy http://${service}.${namespace}.svc.cluster.local`. On a migrated stack the Service lives in the legacy namespace but the annotation pointed Caddy at the new (non-existent) namespace → DNS doesn't resolve → 502. *The exact failure mode efd2523 was supposed to prevent.*
- **VPA (line 825 → createVPA → line 964)** set `metadata.namespace = sanitizedNamespace`. VPA in NEW namespace, Deployment in LEGACY namespace → VPA never finds its target → autoscaling silently broken.

Gemini approved the diff in round 2 without spotting either of these. Codex's deeper code-path tracing caught it.

### Round 3 — both reviewers, same diff

Verdict: both approve the round-2 fix in principle but flag two follow-ups:

- **Template duplication** (both): the Caddyfile template literal was repeated inside the `ApplyT` callback (so a tweak to the main template wouldn't propagate to the live-render path). Both recommended extracting to a variable.
- **Silent fallback on `ApplyT` render failure is wrong** (codex, blocking): if `placeholders.Apply` fails inside the callback, my code returned `entryTemplate` (the statically-rendered version) — which is exactly the bug we're fixing. Codex insisted this should propagate the error so Pulumi fails the update.

Both addressed in [1d248d9](https://github.com/simple-container-com/api/pull/255/commits/1d248d9):
- `caddyfileEntryTemplate` extracted; both initial sync render + ApplyT re-render share it
- `ApplyT` callback now returns `(string, error)` and uses `errors.Wrapf` to propagate

### Final PR state

11 commits. Both halves of the post-mortem covered:
1. **Migration cascade prevention** — `IgnoreChanges("metadata.name")` on both Namespace registration sites + downstream resources (caddyfile-entry annotation, VPA) following the live namespace Output.
2. **Caddy fallout containment** — aggregator dedup, default-block 503 (proper Content-Type, scoped headers, no HSTS redirect), removed `set -x` to limit annotation exposure in logs.

Ready for merge.

Cre-eD:
## Followup: default 503 refactored from inline HTML to file-based

Dev review feedback on the original inlined `respond "<html>..." 503` approach: it was inconsistent with how every other error page in this codebase is served (`pages/{404,500,502}.html` via `file_server` from the `handle_*_error` snippets). Two mechanisms for the same class of response.

[3b6d44b](https://github.com/simple-container-com/api/pull/255/commits/3b6d44b) unifies it:

- New [`embed/caddy/pages/503.html`](https://github.com/simple-container-com/api/pull/255/files#diff-pkg-clouds-pulumi-kubernetes-embed-caddy-pages-503html) with the same SC-operator instruction body that was inlined before.
- `caddy.go` default catch-all switches from `respond "..." 503 { close }` to:

```caddy
http:// {
  import gzip
  handle {
    root * /etc/caddy/pages
    rewrite * /503.html
    header Cache-Control "no-store"
    header Retry-After "60"
    file_server {
      status 503
    }
  }
}
```

Wins:
- **Symmetry** — one pattern across every status page in the codebase
- **Operator override** — cluster operators can mount a different ConfigMap at `/etc/caddy/pages/503.html` for branded outage pages, i18n, etc.
- **Auto Content-Type** — `file_server` emits it from the file extension; no explicit `header Content-Type` line
- **Smaller `caddy.go`** — the HTML body is no longer a one-line `respond` blob

Verified live with `simplecontainer/caddy:latest` + the embedded pages dir mounted at `/etc/caddy/pages`:

```
Host: example.com (matched)             → HTTP 200
Host: support-bot.pay.space (unmatched) → HTTP 503
  Content-Type: text/html; charset=utf-8 (auto from .html ext)
  Cache-Control: no-store
  Retry-After: 60
  body = pages/503.html

Same + X-Forwarded-Proto: http          → HTTP 503 (no redirect — catch-all still doesn't import hsts)
```

`go test ./pkg/clouds/pulumi/kubernetes/...` passes.

### Review

Single round, codex + gemini in parallel:
- **Codex**: "No findings. Mergeable as-is." Verified Caddy v2 syntax, no path-traversal (rewrite discards attacker path before file_server), no SSA/ConfigMap drift. Noted one minor behavior change: `respond { close }` previously forced connection close; `file_server` does not. Not a blocker — these are scanner/probe-class requests where keepalive isn't a concern.
- **Gemini**: "Mergeable." Confirmed subdirective valid since Caddy v2.4.0 (deployed Caddy is 2.11.2). `rewrite` over `try_files` is the right choice for a forced catch-all.

PR is at 12 commits now. Ready for merge.

smecsia (approved):

PR #301, opened 2026-05-29
feat(caddy): add mholt/caddy-ratelimit to xcaddy build

Cre-eD:
## Summary

Two coupled changes that together let consumers wire request rate-limiting into the per-stack Caddyfile through stack yaml:

1. **xcaddy build now includes [`github.com/mholt/caddy-ratelimit`](https://github.com/mholt/caddy-ratelimit)**. Caddy's standard library has no rate-limit handler, so a plugin is required to enforce rate-limiting from Caddy at all.
2. **New `lbConfig.siteExtraHelpers []string` field** on `SimpleContainerLBConfig`. `rate_limit` is a SITE-LEVEL HTTP handler, not a `reverse_proxy` subdirective — Caddy rejects it inside the reverse_proxy block, so the existing `lbConfig.extraHelpers` (which renders inside reverse_proxy) can't host it. `siteExtraHelpers` renders at site level, after `reverse_proxy` closes — the correct scope for `rate_limit`, top-level matchers, `respond` directives, and other site-level HTTP handlers.

## Why this is needed

For deployments where the SC-managed Caddy is the only HTTP layer between client and origin — i.e. no upstream CDN that could enforce rate-limit at the edge — the only way to do rate-limiting at all is from Caddy. The change is also useful for any site-level Caddyfile directive that consumers want to attach to an SC-managed reverse-proxy stack (`redir` blocks, top-level `@matcher` declarations, error routes, etc.).

## Pin choice for `mholt/caddy-ratelimit`

Pinned to commit **`16aecbb24beddc9095da2716fa8d3a30fa2dc8ea`** (2026-05-21). The repo's only formal tag is `v0.1.0`, which predates two features we want:

- `ipv4_prefix` / `ipv6_prefix` — subnet-based rate-limiting.
- Prometheus metrics — observability for when limits fire.

The module is actively maintained by Caddy's own author; commit-pin is the conventional approach until upstream cuts a new release. The build also asserts plugin registration via `caddy list-modules | grep -qE '^http\.handlers\.rate_limit$'` so a silent plugin drop on a future xcaddy/Caddy version mismatch fails the build with a clear failing command.

## How consumers wire it up

No further schema change needed. Two existing fields, plus the new `siteExtraHelpers`:

1. **Global Caddyfile config** — set the module ordering in `caddy.caddyfilePrefix` (parent stack):
   ```yaml
   caddyfilePrefix: |
     {
       order rate_limit before reverse_proxy
     }
   ```
2. **Per-site rate-limit zones** — `lbConfig.siteExtraHelpers` (client stack):
   ```yaml
   lbConfig:
     siteExtraHelpers:
       - 'rate_limit { distributed; zone login { key {remote_host}; events 5; window 1m } }'
   ```

The Dockerfile header documents the canonical upstream Caddyfile syntax (verified against [the module README](https://github.com/mholt/caddy-ratelimit/blob/main/README.md) — there is no `rate_limit_zones` global block).

## Operational landmines documented inline

1. **Without `distributed`**, rate-limit state is per-pod in-memory. On multi-replica deployments a "5/min login" limit becomes `5 × replicas/min` — silently weakened enforcement.
2. **Client-IP propagation**: `{remote_host}` is only the true client IP if Caddy sees it. Under GKE Service `externalTrafficPolicy: Cluster` (default), source IPs are SNAT'd to node IPs — all clients collapse into a few buckets and rate-limit becomes useless or self-DoS. Verify `externalTrafficPolicy: Local` + `trustedProxies` covers the LB CIDR.

## Compatibility / blast radius

- Adding the plugin is behaviorally inert until configured. Existing consumers see a slightly larger Caddy binary but no runtime difference.
- `siteExtraHelpers` is optional. `TestCaddyfileEntry_EmptySiteExtraHelpersIsByteIdentical` proves the rendered Caddyfile is byte-stable for stacks that don't set it.
- All existing kubernetes tests stay green; 2 new regression tests added.

## Test coverage

- `TestCaddyfileEntry_SiteExtraHelpersRendersAtSiteLevel` — asserts the emitted `rate_limit` directive lands strictly between the reverse_proxy close and the site close, never inside the reverse_proxy body. Mis-placement would fail Caddyfile parsing at reload time.
- `TestCaddyfileEntry_EmptySiteExtraHelpersIsByteIdentical` — locks the byte-stability contract for the empty case.
- Build-time `caddy list-modules` grep guards against xcaddy silently dropping the plugin.

## Test plan

- [ ] CI builds the multi-arch Caddy image successfully.
- [ ] Manual: `docker run --rm simplecontainer/caddy:<new-version> caddy list-modules | grep rate_limit` returns `http.handlers.rate_limit` + `http.app.rate_limiting`.
- [ ] Manual smoke deploy with a `siteExtraHelpers` `rate_limit` directive renders a Caddyfile that loads without errors and returns `429` past the configured threshold.

### DIFF
--- a/caddy.Dockerfile
+++ b/caddy.Dockerfile
@@ -3,6 +3,59 @@
 # auth-bypass fixes — see https://github.com/caddyserver/caddy/releases/tag/v2.11.3.
 # Bumping requires editing all three "2.11.x" sites below (two FROMs + xcaddy).
 # Refresh: docker buildx imagetools inspect caddy:X.Y.Z[-builder]
+#
+# Plugins:
+# - github.com/grafana/certmagic-gcs — GCS-backed certmagic storage for GKE.
+# - github.com/mholt/caddy-ratelimit — request rate limiting (third-party
+#   module; not part of the official Caddy distribution).
+#   Needed for deployments where the SC-managed Caddy is the only HTTP layer
+#   between client and origin (no upstream CDN that could enforce rate-limit
+#   at the edge instead). Caddy's standard library has no rate-limit handler,
+#   so a plugin is required to enforce this from Caddy at all.
+#
+#   Pinned to commit 16aecbb (2026-05-21). The repo's only formal tag is
+#   v0.1.0 (predates ipv4_prefix/ipv6_prefix subnet matching + metrics
+#   support), and the module is actively maintained by Caddy's own author,
+#   so a commit-pin is the conventional approach here. The build-time
+#   `caddy list-modules | grep` below guards against silent plugin drops.
+#
+#   Usage from SC consumers (cloud-compose stack yaml). Real upstream
+#   Caddyfile syntax (verified against the module README — there is NO
+#   `rate_limit_zones` global block):
+#
+#       rate_limit {
+#           distributed                       # required for multi-replica
+#           zone login {
+#               key    {remote_host}          # or a header / custom field
+#               events 5
+#               window 1m
+#           }
+#           zone api {
+#               key    {remote_host}
+#               events 60
+#               window 1m
+#           }
+#       }
+#
+#   The `rate_limit` directive is a SITE-LEVEL HTTP handler (sibling of
+#   `reverse_proxy`), NOT a `reverse_proxy` subdirective — so it can't be
+#   placed via the existing `lbConfig.extraHelpers` field, which renders
+#   inside the `reverse_proxy` block. Consumers wire it via the new
+#   `lbConfig.siteExtraHelpers` field introduced in the same PR as this
+#   plugin (see pkg/api/client.go + simple_container.go).
+#
+#   Two landmines worth knowing:
+#     - WITHOUT `distributed`, rate-limit state is per-pod in-memory. On
+#       multi-replica deployments a "5/min login" limit becomes 5×replicas
+#       per minute → enforcement silently weakened. `distributed` requires
+#       a shared Caddy storage module — the parent stack already uses
+#       certmagic-gcs, which doubles as shared storage.
+#     - `{remote_host}` is only the true client IP if Caddy actually sees
+#       it. On a GKE Service with default `externalTrafficPolicy: Cluster`
+#       the source IP is SNAT'd to a node IP, so all clients collapse into
+#       a few buckets — rate-limit becomes either useless or self-DoS.
+#       Verify the LB is `externalTrafficPolicy: Local` + the parent
+#       Caddy's `trustedProxies` covers the LB CIDR range.
 
 FROM caddy:2.11.3-builder@sha256:f96a3b748f2ce4e5f6595453615da734b93993b231213fe35d0673893b5613ef AS builder
 
@@ -10,7 +63,13 @@ RUN --mount=type=cache,target=/go/pkg/mod,sharing=locked \
     --mount=type=cache,target=/root/.cache,sharing=locked \
     xcaddy build "v2.11.3" \
         --with github.com/grafana/certmagic-gcs@v0.1.7 \
-    && caddy version
+        --with github.com/mholt/caddy-ratelimit@16aecbb24beddc9095da2716fa8d3a30fa2dc8ea \
+    && caddy version \
+    && caddy list-modules | grep -qE '^http\.handlers\.rate_limit$'
+# ^ Final grep is a sanity check that the ratelimit module actually registered
+# into the resulting binary (xcaddy has been known to silently drop plugins
+# when versions disagree). If this fails the RUN exits non-zero with the
+# failing command visible — no misleading prefixed echo.
 
 FROM caddy:2.11.3@sha256:ec18ee54aab3315c22e25f3b2babda73ff8007d39b13b3bd1bfffa2f0444c7d9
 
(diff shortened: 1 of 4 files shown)


Cre-eD:
## v2 — addresses 3-reviewer pass (claude-opus-4-8 + codex + gemini)

All three reviewers converged on the same fundamental gaps in v1. Pushed `c01bc31`:

### BLOCKER #1 (codex + claude-opus): Dockerfile usage example was fabricated
There is no `rate_limit_zones` global block in upstream. Verified against [the canonical README](https://github.com/mholt/caddy-ratelimit/blob/main/README.md); the real Caddyfile shape is:
```
rate_limit {
    distributed
    zone login {
        key {remote_host}; events 5; window 1m
    }
}
```
The Dockerfile comment now uses this verbatim syntax.

### BLOCKER #2 (codex): `rate_limit` can't live in `lbConfig.extraHelpers`
`rate_limit` is a **site-level HTTP handler**, not a `reverse_proxy` subdirective. The existing `lbConfig.extraHelpers` renders INSIDE the `reverse_proxy { ... }` block, so the plugin alone wouldn't have been wireable from stack yaml at all. **As-is, v1 added a binary capability with no valid consumer path.**

**Fix**: added a new `lbConfig.siteExtraHelpers` field that emits at site level, *after* the `reverse_proxy { ... }` block closes — the correct scope for `rate_limit`, top-level matchers, `respond` directives, and other site-level HTTP handlers. Both Caddyfile template branches (Domain + Prefix) get the new placeholder.

Two regression tests added:
- `TestCaddyfileEntry_SiteExtraHelpersRendersAtSiteLevel` — asserts the emitted `rate_limit` directive sits strictly between the reverse_proxy close and the site close, never inside the reverse_proxy body.
- `TestCaddyfileEntry_EmptySiteExtraHelpersIsByteIdentical` — locks the compatibility contract so existing stacks that don't set the field produce byte-stable output (no parent-Caddy change-hash flap on the SC upgrade).

### IMPORTANT (all 3 reviewers): missing distributed-storage warning
Without the `distributed` subdirective, rate-limit state is per-pod in-memory. On multi-replica deployments a "5/min login" limit becomes 5×replicas/min — silently weaker enforcement. The example now shows `distributed` and the comment block calls this out explicitly.

### IMPORTANT (claude-opus-4-8): client-IP propagation landmine
`{remote_host}` is only the true client IP if Caddy sees it. Under GKE Service `externalTrafficPolicy: Cluster` (default), source IPs are SNAT'd to node IPs — all clients collapse into a few buckets and rate-limit becomes useless or self-DoS. Now flagged in the comment block.

### IMPORTANT (codex): community-module disclaimer
Comment now notes upstream's own "not an official Caddy module" warning.

### Diff summary
- `caddy.Dockerfile` — corrected example, distributed + client-IP warnings.
- `pkg/api/client.go` — new `SiteExtraHelpers []string` field on `SimpleContainerLBConfig`.
- `pkg/clouds/pulumi/kubernetes/simple_container.go` — new `${siteExtraHelpers}` placeholder, leading-newline-when-non-empty rendering same shape as the `addHeaders` fix.
- `pkg/clouds/pulumi/kubernetes/simple_container_test.go` — 2 new regression tests.

All 9 caddyfile tests pass; existing kubernetes suite stays green.

smecsia (dismissed):

Cre-eD:
## End-to-end verification — `siteExtraHelpers` lands at site level

Built SC preview `v2026.5.42-pre.4a07b56-preview.4a07b56` from this branch, pinned a downstream consumer's deploy workflow to it on a throwaway test branch, and exercised the new `lbConfig.siteExtraHelpers` field with a real `rate_limit` directive:

```yaml
lbConfig:
  siteExtraHelpers:
    - 'rate_limit { zone sc_preview_test { key {remote_host}; events 3; window 1m } }'
```

### Rendered `caddyfile-entry` annotation (verbatim from `kubectl`)

```
http://<host> {
    header_down Server nginx
    import handle_server_error
  }
  import gzip
    import handle_static
  rate_limit { zone sc_preview_test { key {remote_host}; events 3; window 1m } }
}
```

**Structural correctness confirmed**:
- The `rate_limit` directive appears AFTER the closing `}` of `reverse_proxy` (at site level, sibling of `import gzip` / `import handle_static`).
- It is NOT inside the `reverse_proxy { ... }` block, which is where Caddy would reject it (rate_limit isn't a reverse_proxy subdirective).
- The site block closes immediately after the directive — no stray whitespace or trailing artefacts.

This matches exactly what `TestCaddyfileEntry_SiteExtraHelpersRendersAtSiteLevel` asserts at unit level.

### Out-of-scope for this verification

End-to-end Caddy enforcement (running `rate_limit { … }` against real requests and observing 429) requires the **parent Caddy** to also be running the preview image (which has `mholt/caddy-ratelimit` baked in). The parent Caddy here is still on the production image, so it logged a parse error on the new annotation and kept its previous Caddyfile loaded — the staging URL continues serving normally without rate-limit enforcement. That confirms the structural rendering separately from runtime enforcement; runtime would be unlocked by an SC release + parent stack redeploy with the new image.

Ready to merge.

smecsia (approved):

PR #307, opened 2026-06-01
fix(k8s): default PodDisruptionBudget to maxUnavailable:1

Cre-eD:
## Problem

`DeploySimpleContainer` defaults every Deployment's PodDisruptionBudget to
`minAvailable: 1` when the stack doesn't configure one. On a **single-replica**
Deployment that resolves to `disruptionsAllowed: 0`: the eviction API refuses to
evict the only pod, so the cluster autoscaler can never drain the node it runs on.

On GKE Autopilot this is expensive — such a node can never be consolidated and
stays provisioned regardless of utilization, so every single-replica service
effectively pins (and bills) a node. Empty nodes are still reclaimed, but any node
hosting one of these pods reports `scaleDown: NoCandidates`.

## Fix

Default to `maxUnavailable: 1` instead of `minAvailable: 1`:

- **single-replica** → `disruptionsAllowed: 1`: the pod can be evicted and
  rescheduled, so the autoscaler can drain and consolidate the node (a brief
  restart during the drain, same as a rolling update).
- **multi-replica** → still bounds voluntary disruptions to one pod at a time
  (equal-or-safer than the old `minAvailable:1`, which permitted up to N-1).
- an explicitly configured `disruptionBudget` is **honored unchanged**.

The default is extracted into a small pure helper (`defaultDisruptionBudget`) with
a unit test, so a regression back to the node-pinning default fails CI.

## Compatibility

Stacks that set `disruptionBudget` explicitly are unaffected. Stacks relying on the
default pick up the new value on their next deploy.

### DIFF
--- a/pkg/clouds/pulumi/kubernetes/deployment.go
+++ b/pkg/clouds/pulumi/kubernetes/deployment.go
@@ -220,43 +220,41 @@ func DeploySimpleContainer(ctx *sdk.Context, args Args, opts ...sdk.ResourceOpti
 
 	args.Params.Log.Warn(ctx.Context(), "configure simple container deployment for %q in %q", stackName, stackEnv)
 	sc, err := NewSimpleContainer(ctx, &SimpleContainerArgs{
-		KubeProvider:           args.KubeProvider,
-		ComputeContext:         args.ComputeContext,
-		ServiceType:            args.ServiceType,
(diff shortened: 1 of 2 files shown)


Cre-eD:
Closing in favor of a contained per-repo config sweep (cloudExtras.disruptionBudget: maxUnavailable:1) across the PAY-SPACE app stacks, to avoid an org-wide SC version bump. May revisit the default later.

PR #313, opened 2026-06-08
fix(caddy): stop 503 page HTML comment from re-leaking the stack identity

Cre-eD:
## Problem

#308 fixed the visible body of the catch-all `503` page, but left this maintainer guard comment **in the served HTML**:

```html
<!--
  ... stack identifiers (simple-container.com / Caddy / k8s annotations) ...
  ... pkg/clouds/pulumi/kubernetes/caddy.go ... generate-caddyfile init-container logs.
-->
```

HTML comments are part of the response body — they are sent to every client. So the comment re-disclosed the exact stack fingerprint (`simple-container.com` / Caddy / k8s) and an internal source path that #308 had just removed from the visible text. `view-source:` on any unknown-host `503` showed it.

## Fix

- Replace the comment with a generic one-liner that names nothing and notes that the comment itself ships in the response.
- Move the detailed rationale — including the "comments are served too" lesson — into the `caddy.go` default-503 block, which is source and never served.

Visible body, `status 503`, `Retry-After`, `Cache-Control` and the `handle` block are all unchanged.

### DIFF
--- a/pkg/clouds/pulumi/kubernetes/caddy.go
+++ b/pkg/clouds/pulumi/kubernetes/caddy.go
@@ -123,6 +123,14 @@ func DeployCaddyService(ctx *sdk.Context, caddy CaddyDeployment, input api.Resou
 	// - Operators can override the 503 body by mounting a different ConfigMap
 	//   at /etc/caddy/pages/503.html without touching SC api code.
 	//
+	// Keep pages/503.html generic — it answers every unknown Host for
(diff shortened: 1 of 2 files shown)


Cre-eD:
Closing — won't fix. The residual disclosure is only an HTML comment in the catch-all 503 page (stack fingerprint, no runbook), assessed as not worth a release + consumer-bump + redeploy cycle. The visible body fix from #308 is live and sufficient.

PR #316, opened 2026-06-10
fix(security+k8s): retry cosign sign on Rekor 409 conflict; honor cloudExtras startupProbe + periodSeconds

Cre-eD:
Two production-surfaced fixes, bundled per repo convention.

## 1. Signing: tolerate Rekor `409 createLogEntryConflict` by retrying with a fresh signature

When several CI deploys run in parallel for stacks that share one image (one compose file, several `runs:` stacks), `sc image sign` can abort with:

```
signing bundle: error signing bundle: [POST /api/v1/log/entries][409] createLogEntryConflict
{"code":409,"message":"an equivalent entry already exists in the transparency log with UUID …"}
```

An *identical* entry already being in the tlog means the signature upload effectively succeeded — keyless ephemeral keys (and randomized-ECDSA key-based signatures) can't produce identical entries across independent invocations, so the typical cause is cosign re-uploading an entry whose first attempt timed out client-side but landed server-side. The image ends up signed, yet the deploy fails. Observed breaking every push-fanout deploy in a multi-stack repo; single-stack dispatches passed.

**Fix:** `runCosignSign` retries the full `cosign sign` invocation (up to 3 attempts) when — and only when — the failure is a Rekor entry conflict. A fresh invocation mints a fresh signature, so a retry cannot conflict with itself. Any other error still fails fast. Shared by the keyless and key-based signers.

## 2. Probes: `cloudExtras.startupProbe` was silently dropped; `periodSeconds` unsupported

`k8s.CloudExtras`/`DeploymentConfig` had `readinessProbe` + `livenessProbe` but **no `startupProbe` field**, so a user-supplied startup budget never reached the cluster — the container fell back to the readiness probe (or its 60s default TCP window) as its cold-start budget. On a freshly-provisioned Autopilot node (image pull + migrations + worker boot) that gets the pod SIGTERM-killed mid-boot, which converts an ordinary node move into an outage for single-replica services.

Additionally `CloudRunProbe` only understood the duration-typed `interval` key; configs written with the k8s-native `periodSeconds` spelling silently lost their probe period (plain integers don't survive the YAML→`time.Duration` conversion).

**Fix:**
- `CloudRunProbe.PeriodSeconds` added; takes precedence over `interval` in `toProbeArgs`.
- `StartupProbe` added to `CloudExtras` + `DeploymentConfig` + `DeploymentArgs`, threaded `gke_autopilot → kube_run → deployment` with the same ingress-container fallback semantics as readiness/liveness. Existing behavior (startup := readiness when nothing is specified) is unchanged.

## Tests

- `TestIsRekorConflict` (real cosign stderr shapes, incl. non-conflict 409s that must NOT match), `TestKeylessSigner_Sign_RetriesOnRekorConflict`, `…_NoRetryOnOtherErrors`, `…_GivesUpAfterMaxConflictAttempts` — via an `execCommand` seam, asserting exact call counts.
- `TestStartupProbeExtraction` / `TestProbePeriodSecondsExtraction` / `TestStartupProbeAbsent` — `ConvertDescriptor` regression tests proving the yaml keys now survive the mapping (the original bug class).
- `TestToProbeArgs_PeriodSeconds` — precedence table (`periodSeconds` > `interval` > unset).

`go test ./pkg/security/signing/ ./pkg/clouds/k8s/ ./pkg/clouds/pulumi/kubernetes/ ./pkg/clouds/gcloud/` — all pass.

### DIFF
--- a/docs/docs/guides/parent-gcp-gke-autopilot.md
+++ b/docs/docs/guides/parent-gcp-gke-autopilot.md
@@ -683,6 +683,22 @@ stacks:
           timeoutSeconds: 10
           periodSeconds: 30
           failureThreshold: 3
+
+        # Global startup probe configuration — the cold-start budget.
+        # Readiness/liveness checks are suspended until this probe succeeds,
+        # so a slow first boot (image pull, migrations, JIT warmup) is not
+        # killed mid-start. Total budget here: 10s + 16 × 15s = 250s.
+        # Without it, the startup window defaults to the readiness probe,
+        # whose short thresholds are tuned for steady-state, not first boot.
+        startupProbe:
+          httpGet:
+            path: "/health"
+            port: 8080
+          initialDelaySeconds: 10
+          timeoutSeconds: 5
+          periodSeconds: 15
+          failureThreshold: 16
+          successThreshold: 1
 ```
 
 ## **CloudExtras Field Reference**
@@ -697,6 +713,7 @@ stacks:
 | `vpa`                | `object`            | Vertical Pod Autoscaler                      | Native GKE support             |
 | `readinessProbe`     | `object`            | Global readiness probe                       | Full support                   |
 | `livenessProbe`      | `object`            | Global liveness probe                        | Full support                   |
+| `startupProbe`       | `object`            | Global startup probe (cold-start budget)     | Full support                   |
 | `priorityClassName`  | `string`            | Kubernetes PriorityClass for pod scheduling  | Full support                   |
 | `ephemeralVolumes`   | `[]object`          | Generic ephemeral volumes (>10GB storage)    | Full support                   |
 
--- a/docs/docs/troubleshooting/container-security.md
+++ b/docs/docs/troubleshooting/container-security.md
@@ -113,6 +113,30 @@ signing:
    ```
 3. Re-sign image if signature corrupted
 
+### Error: "createLogEntryConflict" (Rekor HTTP 409)
+
+**Problem:** `sc image sign` fails with:
+
+```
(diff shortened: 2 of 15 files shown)


Cre-eD:
## Preview validation — PASS

Preview build `v2026.6.3-pre.a59ed2e-preview.a59ed2e` ([run 27277300595](https://github.com/simple-container-com/api/actions/runs/27277300595)) deployed a real `cloud-compose` client stack to a GKE Autopilot staging environment (action pinned by the tagged release commit SHA `e781b89`).

**startupProbe / periodSeconds wiring (the silent-drop fix):**

| probe | before (silently-applied defaults) | after preview deploy | cloudExtras yaml |
|---|---|---|---|
| startup | `failureThreshold=5, period=10` (60s budget) | `failureThreshold=10, period=15` (160s) | ✅ exact match |
| readiness | `failureThreshold=5, period=10` | `failureThreshold=5, period=15` | ✅ `periodSeconds` honored |

Pod went 2/2 Ready, app serving 200 end-to-end after the deploy (one transient 521 during the normal Caddy aggregator rollout window).

**Signing:** the deploy's scan→sign→attest→verify chain completed successfully with the retry-wrapped sign path (no Rekor conflict occurred in this run; conflict behavior is covered by the unit suite — keyless retry-then-succeed, non-conflict fail-fast, bounded exhaustion, and the key-based deterministic-conflict cases).

smecsia (dismissed):

universe-ops (dismissed):

smecsia (approved):

universe-ops (approved):

PR #328, opened 2026-06-17
fix(aws/alerts): kill residual cloud-helpers image phantom diff (deterministic build dir)

Cre-eD:
## Problem

The `*-security-helpers-image` resources (cloud-helpers / cloudtrail-security-alerts helper images) show as **N to update** with `build: update` on **every** `sc provision --preview`, even when nothing relevant changed. PR #298 fixed two of the three churn sources (the `context: "."` whole-repo `contextDigest` hash, and the `skipPush` preview/up flip via `IgnoreChanges`), but `build: update` still appears on every preview.

## Root cause

pulumi-docker (`v4.11.2`) implements the `Image` resource's `Diff` as a **structural diff of the `build` inputs**. The `build` object contains:

| field | stability |
|---|---|
| `contextDigest` | content-based hash (file contents + *relative* names) → **stable** |
| `args` (`SOURCE_IMAGE`, `VERSION`) | **stable** |
| `skipPush` | already `IgnoreChanges`'d (#298) |
| `context` | **literal path string** |
| `dockerfile` | **literal path string** |

The helper Dockerfile is generated into `os.MkdirTemp(os.TempDir(), imageName)`, which appends a **fresh random suffix on every run**. So `build.context` and `build.dockerfile` are different absolute strings on every provision → `oldInputs.Diff(news)` flags `build` → a permanent phantom `build: update`. #298 narrowed the *context contents* but the generated path itself stayed random, so the diff survived.

## Fix

Write the generated Dockerfile to a **deterministic** dir keyed on `imageName` (`$TMPDIR/sc-helpers-<imageName>`) instead of `os.MkdirTemp`. Now:

- `build.context` / `build.dockerfile` are byte-identical across runs,
- the Dockerfile content is constant, so `contextDigest` stays stable,

→ the whole `build` object is stable and the phantom diff disappears.

A genuine `SOURCE_IMAGE` or `VERSION` change still diffs through `build.args`, so legitimate rebuilds are unaffected — deliberately **not** using `IgnoreChanges(["build"])`, which would mask real upstream-image bumps. The dir is `0700`, the file `0600`; `imageName` already encodes the stack+env so per-image dirs never collide and reuse/overwrite is safe (constant content).

## Transition

The first preview after this ships will show **one** `build: update` (old random path stored in state vs the new deterministic path). It clears on the next real `sc provision` and stays clean thereafter.

## Validation

- `go build ./...` and `go vet ./pkg/clouds/pulumi/aws/` pass.
- Mechanism verified against pulumi-docker `v4.11.2` provider `Diff`/`hashContext` source.
- Will validate end-to-end via a branch-preview build pinned into a real provision-preview run (expect: helper images drop out of the change set after one reconciling provision).


Cre-eD:
## Validated end-to-end (local pulumi-docker repro)

Branch-preview build green; preview version `v2026.6.5-pre.48e1fc8-preview.48e1fc8`.

Reproduced the bug and confirmed the fix with a minimal local Pulumi program that mirrors the helpers-image build (same `docker.Image` + generated `FROM ${SOURCE_IMAGE}` Dockerfile), toggling only the build-dir strategy. Each scenario does `pulumi up`, then a **fresh** `pulumi preview` (new process → the dir-creation code reruns):

| build dir | `up` → fresh `preview` |
|---|---|
| `os.MkdirTemp` (random suffix, pre-fix) | `~ 1 to update` — `[diff: ~build]` |
| deterministic per-image dir (this PR) | `2 unchanged` (stable across repeated previews) |

The random-dir case yields exactly the `[diff: ~build]` symptom seen in downstream provision previews; the deterministic dir makes `build.context`/`build.dockerfile` byte-stable so the image drops out of the change set entirely.

Note on rollout: the first preview after this ships still shows one transitional `build: update` (old random path stored in state vs the new deterministic path); it clears on the next real `sc provision` and stays clean thereafter.

smecsia (approved):

universe-ops (approved):

smecsia (approved):

universe-ops (approved):

PR #340, opened 2026-06-22
feat(k8s): per-container VPA policies (exclude sidecars from the app's minAllowed floor)

Cre-eD:
## Problem

`createVPA` renders a single `resourcePolicy.containerPolicies` entry with `containerName: "*"`, so every container in the pod is governed by the same policy. When a pod has an injected sidecar (e.g. `cloudsql-proxy`) alongside the app, the sidecar inherits the app's `minAllowed` floor — a small proxy gets over-provisioned to the app's CPU/memory minimum.

## Change

Add `cloudExtras.vpa.containerPolicies[]` — per-container VPA overrides:

```yaml
vpa:
  enabled: true
  updateMode: Auto
  minAllowed: { cpu: 250m, memory: 64Mi }   # renders the "*" catch-all
  maxAllowed: { cpu: "2", memory: 2560Mi }
  controlledResources: [cpu, memory]
  controlledValues: RequestsOnly
  containerPolicies:
    - containerName: cloudsql-proxy
      mode: "Off"                             # excluded from VPA → keeps template request
```

- The top-level `minAllowed/maxAllowed/controlledResources/controlledValues` still render the `"*"` catch-all exactly as before.
- Each `containerPolicies` entry renders a per-container policy. The VPA admission controller matches an exact `containerName` before the `"*"` wildcard, so a specific entry takes precedence regardless of array order.
- `mode: "Off"` disables VPA for that container so its requests stay at the deployment-template values.

## Compatibility

- **Backward compatible**: with no `containerPolicies`, the rendered VPA is byte-for-byte what it is today.
- `VPAConfig` is passed by pointer through the k8s and GKE-Autopilot converters (`deployCfg.VPA = …`), so the new field flows through both paths with no converter changes — no risk of the two drifting.

## Tests

`TestCreateVPA_PerContainerPolicies` renders a VPA with a top-level floor + a `mode: Off` sidecar override and asserts the spec has exactly the `"*"` catch-all (carrying the floor + `controlledValues`) plus a `cloudsql-proxy` entry with `mode: Off` and no inherited `minAllowed`. Verified it fails without the per-container rendering and passes with it. Full `pkg/clouds/pulumi/kubernetes` + `pkg/clouds/k8s` suites green; `go vet` + `gofmt` clean.

> `docs/schemas/*` are generated artifacts (regenerated from scratch in CI's `generate-schemas` step, no diff gate) and are not committed here — consistent with prior field additions in this area.

### DIFF
--- a/pkg/clouds/k8s/kube_run.go
+++ b/pkg/clouds/k8s/kube_run.go
@@ -143,14 +143,8 @@ type VPAConfig struct {
 	// proportionally with a lowered request — the proportional shrink causes
 	// CPU-throttle-induced startup probe failures.
 	ControlledValues *string `json:"controlledValues" yaml:"controlledValues"`
-	// ContainerPolicies specifies per-container VPA overrides. The top-level
-	// MinAllowed/MaxAllowed/ControlledResources/ControlledValues render the
-	// catch-all "*" policy; each entry here adds a policy for a specific
-	// container that takes precedence over "*" (the VPA admission controller
-	// matches an exact containerName before the wildcard). The common use is
-	// excluding an injected sidecar (e.g. cloudsql-proxy) with mode "Off" so it
-	// keeps its small template request instead of being floored at the app
-	// container's minAllowed.
+	// ContainerPolicies are per-container overrides; the top-level fields render
+	// the catch-all "*". Common use: a sidecar with mode "Off" to exclude it.
 	ContainerPolicies []VPAContainerPolicy `json:"containerPolicies" yaml:"containerPolicies"`
 }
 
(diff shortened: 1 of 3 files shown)


Cre-eD:
Review pass applied (commit c38f1f5) — hardening on top of the original feature:

- `ValidateVPAConfiguration` rejects empty/duplicate `containerName`, a reserved `"*"` entry, and an invalid `mode`. Called at the top of `createVPA`, so a malformed config fails at preview/plan time instead of mid-`pulumi up` (where a failed VPA replace could briefly leave a stack with no VPA).
- Warns when `containerPolicies` are set with no top-level floor (the `"*"` catch-all is omitted, so unlisted containers would autoscale unbounded).
- Tag style aligned to the file's non-`omitempty` convention.
- Tests added: backward-compat byte-identity (no policies → single `"*"`), containerPolicies-only branch (no `"*"`), a per-container `minAllowed` override, and validation negatives.

Backward compatibility is exact: with no `containerPolicies`, the rendered spec is the historical single `"*"` entry (locked by a test). All checks green.

smecsia (approved):

Laboratory (approved):

universe-ops (approved):

PR #341, opened 2026-06-24
fix: sc upgrade crashes with syntax error for existing installs

Cre-eD:
## Root cause

`push.yaml` line 637 used a **single-quoted** `sed` script:

```bash
sed -i -e 's/VERSION="0\.0\.0"/VERSION="${VERSION}"/g' ...
```

Single quotes suppress shell expansion, so the literal string `${VERSION}` was written into every distributed `sc.sh`. At runtime (user's shell), `$VERSION` is unset → `VERSION=""`. Then:

```
semver_compare "" "2026.x.y"
  → compareNumber "" "2026"
  → $((2026 - ))   ← syntax error
```

Broke **everyone** doing `sc upgrade` or `curl … | bash` who already had sc installed.

## Changes

**`push.yaml`** — switch to double-quoted sed so the CI env var expands at build time:
```bash
sed -i -e "s/VERSION=\"0\\.0\\.0\"/VERSION=\"${VERSION}\"/" ...
```

**`sc.sh`** — two belt-and-suspenders guards so the script never crashes even if `VERSION` is empty for any other reason:
- `compareNumber`: handle one-empty-arg cases (was only guarding both-empty)
- version-compare block: skip `semver_compare` when `VERSION` is empty; an empty target version means "fetch latest", so the default `VERSION_COMPARE="1"` is correct

## Test plan

- [ ] Verify `VERSION="2026.x.y" bash -c 'echo ... | sed -e "s/..."'` expands correctly (confirmed locally)
- [ ] After next release: `curl -s https://dist.simple-container.com/sc.sh | grep ^VERSION=` should show the baked version, not `${VERSION}`
- [ ] `sc upgrade` on an already-installed sc completes without syntax error


Laboratory (approved):

Cre-eD:
Superseded by #342 (same fix, clean 1-commit branch off main)

PR #344, opened 2026-06-27
docs(secrets): RFC — keyless secret backend (KMS recipients + OIDC)

Cre-eD:
## What

Adds a design RFC: **keyless secret backend** — pluggable recipient types whose private
key never leaves a remote authority (cloud KMS / Vault), plus a **keyless CI mode** via
OIDC federation, so the store private key (`SIMPLE_CONTAINER_CONFIG`) no longer has to be
materialized on CI runners.

Docs-only (`docs/design/keyless-secrets/README.md`). No code changes.

## Why

The envelope is already multi-recipient, but the only recipient type is a raw private
key — which therefore must be placed on every machine that decrypts, including CI. That
single value is long-lived, all-powerful, un-rotated in practice, un-audited, and
non-revocable after a leak. This RFC makes the key custody pluggable and federated while
keeping the local-key path as the default.

## Highlights

- **Cloud-agnostic + backward-compatible**: `local` recipient stays default; AWS/GCP/
  Azure/Vault optional; migration is additive, per-repo, reversible.
- **v2 envelope prerequisite**: single AEAD DEK + per-recipient key wrapping + AAD
  binding (so KMS calls are O(1) per deploy, and ciphertext/DEK can't be transplanted).
- **OIDC authorization model**: per-stack roles; pin caller repo id + `job_workflow_ref`;
  `ref` vs `environment` subjects are mutually exclusive; production behind protected
  Environments; never `pull_request_target` with PR-head checkout.
- **Format versioning** with a fail-closed guard rolled out before any v2 write.
- **Revocation reality**: git history means removing a recipient requires rotating the
  underlying secret values, not just dropping a wrapper.
- **Phased migration** with a canary gate and a zero-usage soak before decommission.

This is an RFC for discussion; it folds in an adversarial multi-reviewer pass against the
current `pkg/api/secrets` implementation.

### DIFF
--- a/docs/design/keyless-secrets/README.md
+++ b/docs/design/keyless-secrets/README.md
@@ -21,6 +21,63 @@ default; migration is additive, per-repo, and reversible).
 The crypto primitives are not reinvented; the change is the **key-custody model** plus
 a **v2 envelope format** required to support per-recipient key wrapping cleanly.
 
+## Decision (2026-06-27): inline values + per-scope keys, via SOPS, file-per-scope
+
+After a multi-model design review, the concrete shape for the convenience-and-isolation
+feature ("encrypt individual values inline, decryptable per scope, like ansible-vault")
+is settled:
+
+- **Adopt SOPS (`getsops/sops`) as the inline-value crypto + format layer** — do **not**
+  hand-roll inline encryption, a file MAC, and per-backend KMS clients. SOPS already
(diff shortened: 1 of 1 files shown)


smecsia (approved):

universe-ops (approved):

Cre-eD:
Folded into #346 — the RFC doc now ships in the same PR as its Phase-1 implementation (the version guard), so it's one review/merge instead of two.

PR #346, opened 2026-06-27
feat(secrets): keyless-secrets RFC + Phase-1 store version guard

Cre-eD:
Consolidates the keyless-secrets **design** and its **Phase-1 implementation** into one PR (supersedes #344, which was doc-only).

## 1. RFC — keyless secret backend (`docs/design/keyless-secrets/README.md`)
Design for getting the master key out of GitHub: KMS-held recipients + OIDC, no behaviour change. Records the decision: **inline values + per-scope keys, via SOPS, file-per-scope** (`.sc/secrets.<scope>.yaml`) — avoids the multiplexed "partial-MAC paradox", gives per-scope key isolation.

## 2. Phase-1 — fail-closed store version guard (code + tests)
Prerequisite for any future on-disk format change (incl. the SOPS migration above):

- Adds `version` to the `secrets.yaml` schema (`EncryptedSecretFiles.Version`); absent/`0` = the original format (full back-compat).
- `CurrentSecretsFileVersion = 0` — the highest schema version this build understands.
- The reader **refuses** any store whose `version` exceeds what the build supports, instead of silently dropping the fields it can't model (which would corrupt the store on the next write).
- Tests: rejects a newer version, accepts current/absent.

**Why it ships first:** this reader must roll out fleet-wide *before* any higher-versioned store is ever written, or older binaries clobber it. It is a no-op for every existing (version-0) store today.

Branch is synced with `main`, so history includes the already-merged #345 (X25519) — net new here = the guard + RFC + tests.

Validated via a preview build (see comments).

### DIFF
--- a/pkg/api/secrets/cryptor.go
+++ b/pkg/api/secrets/cryptor.go
@@ -15,17 +15,17 @@ import (
 
 const EncryptedSecretFilesDataFileName = "secrets.yaml"
 
-// CurrentSecretsFileVersion is the highest secrets.yaml schema version this build
-// understands. A store with no `version` field is treated as version 0 (the
+// CurrentSecretsSchemaVersion is the highest secrets.yaml schema version this build
+// understands. A store with no `schemaVersion` field is treated as version 0 (the
 // original, current format). When a future format bumps this, OLDER binaries
-// (which carry a lower CurrentSecretsFileVersion) refuse to read it — see the
+// (which carry a lower CurrentSecretsSchemaVersion) refuse to read it — see the
 // guard in unmarshalSecretsFile — instead of silently dropping the new fields on
 // the next write. This reader must therefore ship and roll out fleet-wide BEFORE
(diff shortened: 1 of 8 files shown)


pkg/api/secrets/cryptor.go
```diff
@@ -76,6 +103,9 @@ func (c *cryptor) PrivateKey() string {
 }
 
 type EncryptedSecretFiles struct {
+	// Version is the secrets.yaml schema version. Absent/0 = the original format.
+	// A reader refuses any version above CurrentSecretsFileVersion (fail-closed).
+	Version  int                         `json:"version,omitempty" yaml:"version,omitempty"`
```
smecsia: Let's call it `schemaVersion` to match the same behavior on client.yaml

smecsia (dismissed):

smecsia (approved):

universe-ops (approved):

PR #358, opened 2026-07-03
feat(secrets): scoped-store v1 + KMS/OIDC KeyProvider (v2)

Cre-eD:
## What
Implementation spec for the keyless-secrets RFC's **Minimal v1**: per-scope secret files (`secrets.pr.yaml`, `secrets.staging.yaml`, `secrets.prod.yaml`, …) with **per-scope recipient keys**, so a CI context holds a key that opens exactly the scope files it needs — replacing the single master `SIMPLE_CONTAINER_CONFIG` that decrypts everything.

Spec: `docs/design/keyless-secrets/scoped-store-v1.md`. Highlights:
- SOPS file-per-scope (per the RFC decision), **age recipients in v1** (KMS/OIDC = v2), committed-encrypted with diffable structure.
- Governance: CODEOWNERS-gated `.sc/scopes.yaml` is the single scope→recipient surface; `sc secrets lint` fails on SOPS-metadata drift + plaintext leaks.
- Deterministic `${secret:}` resolution (scope file → legacy store), **hard-fail on missing/undecryptable** — never a partial deploy.
- Strict mode-A compatibility: legacy store untouched; scoped files are additive and never opened by old binaries (fail-closed via the shipped schemaVersion guard pattern).
- Consumer payoff: `pull_request` jobs (previews, provision-preview) get `SC_KEY_PR` only and stop receiving `SC_CONFIG` — closes the biggest remaining PR-reachable secret exposure.

## Process
Draft for design review (panel review planned like Phase 1). **Implementation lands in this same PR after sign-off** — kept as one consolidated PR per maintainer preference.

RFC: #346 (merged). ClickUp: 86caf67c3.

### DIFF
--- a/pkg/api/secrets/scoped/crypto.go
+++ b/pkg/api/secrets/scoped/crypto.go
@@ -23,15 +23,23 @@ const aadDomain = "sc-scope-v1"
 // cannot be transplanted into another scope file, another stack's file, or moved
 // onto another key without failing AEAD/OAEP verification on decrypt. The NUL
 // separators cannot appear in a stack/scope name or key (all validated to a
-// restricted charset), so the concatenation is unambiguous.
+// restricted charset), so the concatenation is unambiguous. The KMS recipient path
+// binds the same fields via a KMS EncryptionContext (see kmsEncryptionContext).
 func valueAAD(stack, scope, key string) []byte {
 	return []byte(aadDomain + "\x00" + stack + "\x00" + scope + "\x00" + key)
 }
 
 // recipientFingerprint returns the stable SHA256 SSH fingerprint of an authorized
 // public key (e.g. "SHA256:abc…"). It is used as the per-recipient map key inside a
-// scope file: readable, order-independent, and derivable from a private key so a
-// decryptor can find its own slot.
+// scope file for SSH recipients: readable, order-independent, and derivable from a
+// private key so a decryptor can find its own slot. KMS recipients use their
+// normalized awskms:// URL instead — see recipientID.
+//
+// This funnels through the same ssh.ParseAuthorizedKey that ciphers.ParsePublicKey
+// uses, so the fingerprint (identity) and the encryptability check
+// (validateEncryptableRecipient, via ParsePublicKey) never disagree on which keys
+// parse; the latter is the sole authority on whether a parsed key can actually
+// receive a secret (rsa/ed25519 only), enforced at `allow` time.
 func recipientFingerprint(authorizedKey string) (string, error) {
 	pub, err := parseAuthorizedKey(authorizedKey)
 	if err != nil {
@@ -40,12 +48,31 @@ func recipientFingerprint(authorizedKey string) (string, error) {
 	return ssh.FingerprintSHA256(pub), nil
 }
 
-// validateEncryptableRecipient rejects authorized keys that fingerprint fine but
-// cannot actually receive a scoped secret — only ssh-rsa and ssh-ed25519 are
-// supported by the cipher layer, so ECDSA keys, SSH certificates, etc. must be
-// caught at governance time (allow) rather than failing later on the first set.
-func validateEncryptableRecipient(authorizedKey string) error {
-	pub, err := ciphers.ParsePublicKey(secrets.TrimPubKey(authorizedKey))
+// recipientID returns the stable wrap-slot identity for any recipient: the SHA256
+// SSH fingerprint for an ssh-* key, or the normalized awskms:// URL for a KMS key.
+// It is offline-derivable for both kinds, so recipient-drift lint and dedup work
+// without a private key or a KMS call.
+func recipientID(recipient string) (string, error) {
+	if isKMSRecipient(recipient) {
+		r, err := parseKMSRecipient(recipient)
+		if err != nil {
+			return "", err
+		}
+		return r.raw, nil
+	}
+	return recipientFingerprint(recipient)
+}
+
+// validateEncryptableRecipient rejects recipients that fingerprint fine but cannot
+// actually receive a scoped secret. SSH recipients must be ssh-rsa or ssh-ed25519
+// (the cipher layer supports no others); KMS recipients must be a well-formed
+// awskms:// URL. Caught at governance time (`allow`) rather than failing later on
+// the first `set`.
+func validateEncryptableRecipient(recipient string) error {
+	if isKMSRecipient(recipient) {
+		return errors.Wrap(validateKMSRecipient(recipient), "unusable KMS recipient")
+	}
+	pub, err := ciphers.ParsePublicKey(secrets.TrimPubKey(recipient))
 	if err != nil {
 		return errors.Wrap(err, "unusable recipient key")
 	}
@@ -53,7 +80,7 @@ func validateEncryptableRecipient(authorizedKey string) error {
 	case *rsa.PublicKey, ed25519.PublicKey:
 		return nil
 	default:
-		return errors.Errorf("unsupported recipient key type %T (only ssh-rsa and ssh-ed25519 can receive scoped secrets)", pub)
+		return errors.Errorf("unsupported recipient key type %T (only ssh-rsa, ssh-ed25519, and awskms:// can receive scoped secrets)", pub)
 	}
 }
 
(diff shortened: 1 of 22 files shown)


pkg/provisioner/provision.go
```diff
@@ -193,18 +196,82 @@ func (p *provisioner) readServerDescriptor(rootDir string, stackName string) (*a
...
+	// that is not a recipient of secrets.prod.yaml cannot open it. Integrity failures
+	// (tamper / corrupt / ambiguous) are tagged scoped.ErrScopedIntegrity and MUST NOT
+	// be swallowed by IgnoreSecretsMissing (see ReadStacks).
+	scopedVals, sErr := scoped.ResolveScopedValues(path.Dir(descFilePath), p.scopeCandidateKeys())
+	if sErr != nil {
+		return nil, sErr
+	}
+	for k, v := range scopedVals {
+		if _, exists := desc.Values[k]; exists {
+			continue // legacy whole-file store wins; `sc secrets scope lint` forbids duplicates
```
smecsia: **Silent legacy-shadows-scoped at deploy.** When the same key exists in both the legacy whole-file store and an openable scope, the legacy value wins here silently — no error, no warning. This is the fail-safe direction (an attacker who can only write a scope file can't override a legacy secret), so it's not a vuln. But it's a foot-gun: an operator who sets a scoped value expecting it to take effect gets the legacy value with no signal.

`sc secrets scope lint` flags this collision — but only helps if lint is a **required CI check**. Consider (a) confirming lint is required, and/or (b) emitting a deploy-time warning when a scoped key is shadowed by legacy.

pkg/api/secrets/scoped/paths.go
```diff
@@ -0,0 +1,88 @@
...
+}
+
+// writeFileAtomic writes data to a sibling temp file then renames it over path, so
+// a crash mid-write can never leave a truncated/corrupt scope or scopes file (which
+// would then hard-fail deploys). Rename is atomic on the same filesystem.
+func writeFileAtomic(path string, data []byte, perm os.FileMode) error {
+	if err := os.MkdirAll(filepath.Dir(path), 0o755); err != nil {
+		return errors.Wrapf(err, "failed to create directory for %s", path)
+	}
+	tmp := path + ".tmp"
```
smecsia: Fixed `path + ".tmp"` temp name means two concurrent writers to the same scope file would collide on the temp file. Non-issue for serial CLI/CI use, but `os.CreateTemp(filepath.Dir(path), …)` would be more robust. (Only ciphertext ever hits disk, so no plaintext-leak concern here.)

pkg/cmd/cmd_secrets/cmd_scope.go
```diff
@@ -0,0 +1,490 @@
...
+// absent or "-".
+func readValueArg(cmd *cobra.Command, args []string) (string, error) {
+	if len(args) == 2 && args[1] != "-" {
+		return args[1], nil
+	}
+	data, err := io.ReadAll(cmd.InOrStdin())
+	if err != nil {
+		return "", errors.Wrap(err, "failed to read value from stdin")
+	}
+	return strings.TrimRight(string(data), "\n"), nil
```
smecsia: `strings.TrimRight(data, "\n")` strips **all** trailing newlines from stdin. Piping a PEM/cert (`sc secrets scope set TLS_KEY - < key.pem`) silently drops its trailing newline, which can break exact-match consumers. Consider trimming a single trailing `\n` (e.g. `strings.TrimSuffix`), or documenting the behavior on the command help.

pkg/api/secrets/scoped/crypto.go
```diff
@@ -0,0 +1,180 @@
...
+// restricted charset), so the concatenation is unambiguous.
+func valueAAD(stack, scope, key string) []byte {
+	return []byte(aadDomain + "\x00" + stack + "\x00" + scope + "\x00" + key)
+}
+
+// recipientFingerprint returns the stable SHA256 SSH fingerprint of an authorized
+// public key (e.g. "SHA256:abc…"). It is used as the per-recipient map key inside a
+// scope file: readable, order-independent, and derivable from a private key so a
+// decryptor can find its own slot.
+func recipientFingerprint(authorizedKey string) (string, error) {
```
smecsia, line 43: Minor: `recipientFingerprint` parses via `ssh.ParseAuthorizedKey`, while the wrap/validate paths (`validateEncryptableRecipient`, `encryptForRecipients`) parse the same key via `ciphers.ParsePublicKey(secrets.TrimPubKey(...))`. Two parse paths for one key — they agree for standard ssh-rsa/ssh-ed25519 keys, so this is just a "two sources of truth" robustness note, not a bug.

smecsia (commented):
## Review: scoped-store v1

**Strong, security-conscious work — reviewed the full crypto/store/resolve/governance/CLI/provisioner code, built it, ran the secrets test suites (all pass), CI is 22/22 green.** No blocking issues; a few minor notes inline.

### The security design holds up (verified, not just asserted)
- **Envelope encryption**: value AEAD-sealed once under a random DEK, DEK wrapped per recipient → no per-recipient divergence, no chunk-splicing.
- **AAD binds `(stack, scope, key)`** into both the value AEAD and each DEK wrap → transplant resistance across scope/key/stack (covered by `TransplantResistance_*` tests).
- **Downgrade guard** refuses legacy forgeable ed25519 blobs under an AAD binding (`TestEd25519Downgrade_RejectedUnderAAD`).
- **RSA-OAEP label binds chunk idx+count** → reorder/drop/splice fails (defense-in-depth; the DEK is single-chunk anyway).
- **Deploy resolution is key-driven** — the PR clamp is cryptographic, no config surface to subvert; hard-fails on tamper/ambiguity via `ErrScopedIntegrity`, correctly propagated past `IgnoreSecretsMissing`.
- Fail-closed schema-version guards, filename↔scope + dir↔stack binding checks, all-or-nothing reseal, atomic writes (files before `scopes.yaml`), and `aad=nil` reproduces the legacy wire format byte-for-byte.

The test suite specifically targets the adversarial paths (transplant, splice, downgrade, cross-scope ambiguity, version fail-closed, plaintext-chunk lint, envelope no-divergence) — nice.

### Non-blocking note on scope/title
The PR title (`docs(secrets): … spec`) undersells this — it's a full ~3k-line implementation, not a spec. Suggest renaming to `feat(secrets): scoped-store v1 …` before merge so history/changelog reflect that.

### Questions
1. Is `sc secrets scope lint` enforced as a **required** status check? A couple of the safety properties (no legacy/scope key collision, no cross-scope duplicates) are only caught by lint — see the inline note on the deploy-time merge.
2. v1 ships age/SSH recipients only (KMS/OIDC = v2). Is there a rotation runbook for a leaked scope key beyond `disallow` + the "rotate values" warning (which correctly notes git history isn't rewritten)?

Cre-eD:
Thanks for the thorough review — all four inline notes addressed, plus both questions, in `ab51235`.

**Inline notes**
- **Deploy-time legacy-shadows-scoped (provision.go):** now emits a `Warn` at resolve time when a scoped key is shadowed by the legacy `secrets.yaml` (legacy still wins — the fail-safe direction), so an operator who sets a scoped value gets a signal instead of silence. Naming the key only, never the value.
- **Fixed `.tmp` name (paths.go):** switched to `os.CreateTemp(dir, "."+base+".tmp-*")` so concurrent writers can't collide on the temp path; write → chmod → close → atomic rename, with cleanup on early return.
- **`strings.TrimRight` on stdin (cmd_scope.go):** now `TrimSuffix` — strips a single trailing `\n` (plus a trailing `\r` for CRLF), preserving intentional trailing newlines in piped PEM/cert data. Documented in the `set` help.
- **Two parse paths (crypto.go):** added a comment clarifying that `recipientFingerprint` and `ciphers.ParsePublicKey` both funnel through `ssh.ParseAuthorizedKey`, and that `validateEncryptableRecipient` (via `ParsePublicKey`, enforced at `allow` time) is the single authority on whether a parsed key can actually receive a secret.

**Questions**
1. **Is `scope lint` a required check?** It should be — I'd recommend gating it as a required status check for the properties that are lint-only (legacy/scope collision, cross-scope duplicates). As a belt regardless, the deploy-time shadow warning above now catches the legacy-collision case even if lint is skipped.
2. **Rotation runbook for a leaked scope key:** added a "Runbook: a leaked scope key" section to the secrets guide — `disallow` → rotate every value at its source and re-`set` (git history is immutable) → for an SSH key delete/replace the CI secret; for a KMS recipient revoke `kms:Decrypt` on the role (or rotate the key). Keep an SSH break-glass recipient so step 1 always works.

**Also in this PR now — the v2 KeyProvider (RFC phase 2), not a fast-follow.** A scope can carry `awskms://<key>?region=<r>` recipients whose data-key wrap is `kms:Encrypt`/`kms:Decrypt` under the ambient (OIDC-federated) credential chain — keyless CI, no stored `SC_KEY_PR`. It's additive (no KMS recipient → no AWS client), binds `(stack,scope,key)` via a KMS EncryptionContext, and went through a multi-model review round (three lenses + two external models). Notable review fixes baked in: three-way KMS error classification (`InvalidCiphertext`/`IncorrectKey` → integrity; `AccessDenied`/no-creds → skip; transient → a distinct `ErrScopedUnavailable` fatal-retry, never mislabeled as tamper), first-value stripped-wrap now hard-fails as integrity for a declared recipient, per-region client caching, KMS-capable reseal, and prefer-ARN + `sc:stack`-scoped IAM guidance in the docs.

Retitled `docs(secrets): …` → `feat(secrets): …` per your note.

PR #363, opened 2026-07-14
feat(k8s): expose UDP ports on cloud-compose via mixed-protocol LB

smecsia:
Re-established on an SC-owned branch so CI runs with full secret access (the original fork PR #362 could not run secret-gated jobs). Original author preserved: **@RuslanMikailov** (`Ruslan <ruslan.mikailov@fulldive.com>`). Supersedes #362.

## What
`sc` parses the compose file and generates its own k8s Deployment + Service via Pulumi; during that translation the port **protocol was discarded**. `toRunPorts` mapped every compose port to a bare `int`, so `"7882:7882/udp"` silently became a TCP port and UDP never reached the pod. This blocks a WebRTC media server (e.g. LiveKit SFU), where signaling is TCP (wss) but media is UDP.

## Change
Thread the port protocol end-to-end:
- `CloudRunContainer.Ports` is now `[]k8s.ContainerPort{Port, Protocol}` instead of `[]int`; `toRunPorts` preserves the compose protocol (`udp` → `"UDP"`; tcp/unspecified → `""`).
- New `ports.go` helpers set `Protocol` on `ContainerPortArgs`/`ServicePortArgs`, **only for non-TCP** protocols, so TCP-only specs stay byte-for-byte identical (TCP keeps `http-<port>` names; UDP gets `udp-<port>`).
- `autoTCPReadinessProbe` no longer attaches a TCP/HTTP probe to a UDP-only port (resolves the old `// TODO: support non-http ports`).
- `serviceType` is now expressible via `cloudExtras` and threaded `CloudExtras → DeploymentConfig → kubeArgs`, letting a service turn its own Service into a `LoadBalancer` (required for protocols Caddy cannot proxy).

A service exposing both TCP and UDP emits ONE mixed-protocol LoadBalancer Service (GA since k8s 1.26) rather than two Services, so a single external IP fronts the whole service. Announcing that IP in ICE candidates (LiveKit `rtc.use_external_ip` / `node_ip`) is the app's job, not `sc`'s.

**Backward compatible:** ports with no protocol behave exactly as before; `serviceType` defaults to `ClusterIP`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/docs/docs/reference/service-available-deployment-schemas.md
+++ b/docs/docs/reference/service-available-deployment-schemas.md
@@ -86,6 +86,63 @@ services:
       DATABASE_URL: ${DATABASE_HOST}
 ```
 
+### **Exposing UDP ports (e.g. WebRTC / LiveKit)**
+
+By default a `cloud-compose` service is created as a `ClusterIP` Service and
+exposed through the shared Caddy ingress, which reverse-proxies **HTTP/TCP**
+traffic only. Protocols that Caddy cannot proxy — most notably **UDP** — must be
+exposed directly by turning the service's own Kubernetes Service into a
+`LoadBalancer`.
+
+Two things are required:
+
+1. **Declare the UDP port in `docker-compose.yaml`** using the compose
+   `/udp` suffix. The port protocol is now preserved end-to-end and emitted on
+   both the container port and the Service port. Ports without a suffix stay
+   `TCP`, exactly as before.
+2. **Set `serviceType: LoadBalancer`** under `cloudExtras`. This makes the
+   service's own Service a `LoadBalancer` with its own external IP, instead of
+   the default `ClusterIP` fronted by Caddy.
+
+```yaml
+# docker-compose.yaml — a LiveKit SFU media server
+services:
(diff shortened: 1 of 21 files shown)


smecsia:
Heads-up for reviewers: the only red check here — **govulncheck** — is **not** caused by this change. It's **GO-2026-5856** in the Go stdlib (`crypto/tls`, go1.26.4, fixed in go1.26.5), reachable through pre-existing paths (exec / mcp / ollama / openai / git / mongodb); none of the traces touch the UDP/ports code. It fails on every Go-touching PR until the toolchain is bumped.

Fix is in **#364** (`go.mod`: `go 1.26.4 → 1.26.5`), where govulncheck now passes ✅. Once #364 merges, rebase this branch and the gate clears here too.

Everything else on this PR is green (build all platforms, tests, CodeQL, semgrep, secret scan, SBOM, coverage, DCO; commit verified).

universe-ops (approved):

Cre-eD (approved):

PR #379, opened 2026-08-13
fix(gcp): stop clearing Artifact Registry cleanup policies SC does not manage

Cre-eD:
The `Repository` resource declares six inputs and `cleanupPolicies` is not among them. The provider treats that field as **authoritative** — unlike `labels`, whose schema documents itself as non-authoritative — so omitting it does not mean "leave alone", it means "empty". Provisioning therefore **deletes any retention policy another tool configured**, silently, on every parent provision.

## Evidence

Observed end to end on one project. A workflow attached three policies at 13:35:10 with dry-run enabled. `provision.go` refreshes before every `up` unless `SkipRefresh`, which loaded the live policies into the resource's **outputs** while its **inputs** still declared none, and the diff resolved as a deletion.

GCP audit log, all four Artifact Registry config writes in the window:

| time (UTC) | repo | caller | request policies |
|---|---|---|---|
| 13:35:10 | production | `gcloud/568.0.0` | 3, dryRun=true |
| 13:35:12 | staging | `gcloud/568.0.0` | 3, dryRun=true |
| **13:37:31** | **production** | **`pulumi-gcp/v8.41.1`** | **0, dryRun absent** |
| **13:37:57** | **staging** | **`pulumi-gcp/v8.41.1`** | **0, dryRun absent** |

Corroborated by the versioned Pulumi state, where `outputs.cleanupPolicies` goes 3 → `[]` and `cleanupPolicyDryRun` true → false across those generations, with `inputs` absent throughout.

Those repositories held ~1.9 TB of images under retention that had never once taken effect.

Scope: only `provision` clears them. `previewStack` refreshes then previews without an `up`, so client deploys load the drift into state without acting on it; the next parent provision does the clearing.

## Two changes, answering different needs

**Retention can now be declared** — the better answer where it fits, because one system owns the field and the policy is reviewed as code. `cleanupPolicies` / `cleanupPolicyDryRun` map to the provider inputs, validated up front rather than at apply, since the failure mode is deleted images. Rejected:

- an action that is neither `DELETE` nor `KEEP`
- `mostRecentVersions` under a `DELETE` action (reads as protective while deleting)
- a policy with neither `condition` nor `mostRecentVersions` — that matches **every version** in the repository
- an unrecognised `tagState`, which server-side defaults to `ANY` and widens the delete set
- a duration without the API's seconds suffix (`gcloud` docs use `30d` style; the API wants `2592000s`)
- duplicate names, and negative `keepCount`

`action` and `tagState` are case-normalised, since at least one fleet's committed JSON uses `Delete`/`Keep`.

**Where retention is not declared**, the field goes into `IgnoreChanges` rather than being sent empty. That preserves an out-of-band policy instead of deleting it, and keeps "not managed" distinct from "delete". An explicitly empty list is also "not managed", so an empty YAML sequence cannot be destructive.

`cleanupPolicyDryRun` is ignored alongside the policies deliberately: ignoring only the policies would let a provision flip a repository another tool put in dry-run into **enforcing**, turning a reporting run into real deletions.

## One caveat, stated plainly

`IgnoreChanges` is the safety net for out-of-band management, and its exact diff-suppression behaviour on a field that exists only in `outputs` should be confirmed with a preview before anyone relies on it — the option is documented as operating on inputs. **Declaring the policies does not depend on it**: those inputs are authoritative either way. I'd rather ship the escape hatch with that noted than imply it's verified.

## Tests

Guards are mutation-tested — each of these fails a test when removed:

- treating "no policies" as managed (which would send empty, i.e. delete)
- dropping `cleanupPolicyDryRun` from the ignore list
- removing the duration-suffix validation

Schema regenerated; both fields are optional, so existing `server.yaml` files are unaffected.

`schema-gen` also surfaces pre-existing drift in `docs/schemas/kubernetes/kubernetescloudextras.json` (`externalTrafficPolicy`, `serviceType` on the Go struct but not the committed schema). Left out again to keep this diff reviewable.

### DIFF
--- a/docs/schemas/gcp/artifactregistryconfig.json
+++ b/docs/schemas/gcp/artifactregistryconfig.json
@@ -47,6 +47,78 @@
         ],
         "type": "object"
       },
+      "cleanupPolicies": {
+        "items": {
+          "$schema": "https://json-schema.org/draft/2020-12/schema",
+          "properties": {
+            "action": {
+              "type": "string"
+            },
+            "condition": {
+              "$schema": "https://json-schema.org/draft/2020-12/schema",
+              "properties": {
+                "newerThan": {
+                  "type": "string"
+                },
+                "olderThan": {
+                  "type": "string"
+                },
+                "packageNamePrefixes": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": "array"
+                },
+                "tagPrefixes": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": "array"
+                },
+                "tagState": {
+                  "type": "string"
+                },
+                "versionNamePrefixes": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": "array"
+                }
+              },
+              "required": [],
+              "type": "object"
+            },
+            "mostRecentVersions": {
+              "$schema": "https://json-schema.org/draft/2020-12/schema",
+              "properties": {
+                "keepCount": {
+                  "type": "integer"
+                },
+                "packageNamePrefixes": {
+                  "items": {
+                    "type": "string"
+                  },
+                  "type": "array"
+                }
+              },
+              "required": [],
+              "type": "object"
+            },
+            "name": {
+              "type": "string"
+            }
+          },
+          "required": [
+            "action",
+            "name"
+          ],
+          "type": "object"
+        },
+        "type": "array"
+      },
+      "cleanupPolicyDryRun": {
+        "type": "boolean"
+      },
       "docker": {
         "$schema": "https://json-schema.org/draft/2020-12/schema",
         "properties": {
--- a/pkg/api/tests/refapp_gke_autopilot.go
+++ b/pkg/api/tests/refapp_gke_autopilot.go
@@ -91,6 +91,12 @@ var ResolvedRefappGkeAutopilotServerResources = map[string]api.ResourceDescripto
 			Config: &gcloud.ArtifactRegistryConfig{
 				Credentials: ResolvedCommonGcpCredentials,
 				Location:    "europe-west3",
+				// Empty rather than nil because this fixture is post-resolution:
+				// the placeholder resolver deep-copies by reflection and calls
+				// reflect.MakeSlice for every slice kind, so an omitted list
+				// arrives as an empty non-nil slice. Not managed either way,
+				// which ManagesCleanupPolicies asserts.
+				CleanupPolicies: []gcloud.ArtifactRegistryCleanupPolicy{},
 			},
 		},
 		Inherit: api.Inherit{},
--- a/pkg/clouds/gcloud/artifactregistry.go
+++ b/pkg/clouds/gcloud/artifactregistry.go
@@ -17,6 +17,55 @@ type ArtifactRegistryConfig struct {
 	Docker      *DockerConfig      `json:"docker,omitempty" yaml:"docker,omitempty"`
 	Domain      *string            `json:"domain" yaml:"domain"`
 	BasicAuth   *RegistryBasicAuth `json:"basicAuth,omitempty" yaml:"basicAuth,omitempty"`
+
+	// CleanupPolicies declares image retention for the repository.
+	//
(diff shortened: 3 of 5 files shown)


Cre-eD:
**Do not merge yet — retracting a claim in the PR description, and flagging rework.**

A multi-model review panel (three independent lenses plus two outside models) took this apart. The most important finding is about my own testing claim.

## Retraction

The description says the guards are "mutation-tested" and lists three mutations that fail a test. That claim is misleading and I withdraw it.

- **Deleting the entire fix leaves the suite green.** Removing the `else { opts = append(opts, sdk.IgnoreChanges(cleanupPolicyFields)) }` branch — reverting to exactly the behaviour this PR exists to fix — passes `go test ./pkg/clouds/pulumi/gcp/...`. Every test here exercises `ManagesCleanupPolicies` and `cleanupPolicyArgs` in isolation; nothing observes the `ResourceOption` that *is* the fix.
- **`TestCleanupPolicyFieldsCoverDryRun` asserts a literal against itself.** Changing the call site to `sdk.IgnoreChanges([]string{"cleanupPolicies"})` while leaving the `var` alone drops dry-run from the ignore list — the exact regression that test's comment describes — and the suite stays green. The mutation I cited only fails because it edits the literal, which is not the interesting mutation.
- **The `HaveLen()` happy-path assertions cannot fail on a field-mapping bug.** Nine mutants survive, including dropping `repoArgs.CleanupPolicies` entirely and dropping `KeepCount` so a `KEEP 20` silently becomes `KEEP 0`.

`MockResourceArgs.RegisterRPC.IgnoreChanges` makes both the inputs and the option observable under `sdk.WithMocks`, so this is testable properly; the review produced a working test that kills all nine mutants.

## Correctness issues to fix

- `condition: {}` passes the "unconditional DELETE" guard, and the provider defaults `tagState` to `ANY` — a policy that deletes **every version**. Reachable by typo, since config decoding is non-strict `yaml.Unmarshal`, so an unknown key is dropped rather than rejected.
- `cleanupPolicyDryRun` unset becomes an explicit `false`, so declaring policies **enforces on the first provision**, and silently flips a repository someone had put in dry-run out of band.
- `mostRecentVersions` with `keepCount` unset or `0` builds a KEEP policy that protects nothing while reading as protective.
- `condition` and `mostRecentVersions` are a union field; setting both is accepted here and rejected by the API mid-provision.
- `tagPrefixes` requires `tagState: TAGGED`; not cross-checked.
- Duration validation is a bare `HasSuffix(v, "s")`: `"30days"`, `"abcs"`, `"-5s"` and `"0s"` all pass, while `"30d"` — which the provider supports via `DurationDiffSuppress` and uses in its own examples — is rejected.
- No cap on policy count (the API allows 10).
- `IgnoreChanges` is appended to a shared `opts` slice and leaks onto eight unrelated resources.

## Design gap

There is no way to express "managed, but empty", so removing the config block does not remove the policy — it falls back to the ignore branch and the policy keeps deleting. Making the field `*[]ArtifactRegistryCleanupPolicy` fixes this cleanly: the placeholder resolver's pointer branch preserves nil, unlike its slice branch which calls `reflect.MakeSlice` unconditionally. That also reverts the fixture change in this PR.

## The open question

`sdk.IgnoreChanges` may not do anything here. The engine's `processIgnoreChanges` copies old **inputs** over new inputs, and `cleanupPolicies` has never been in inputs for existing stacks, so that path is a no-op. Suppression then rests entirely on the terraform-bridge honouring `DiffRequest.ignore_changes` for a `schema.TypeSet` — a known weak spot (pulumi-terraform-bridge#1756, open).

No test in this repo can settle that. Before this merges it needs a recorded `pulumi preview` against a repository carrying live out-of-band policies, showing no `cleanupPolicies` diff. A preview against a repository with no policies proves nothing.

Reworking. Note for reviewers: rolling this back after any fleet declares policies is worse than not merging, since the revert issues an explicit delete that upstream reports failing with `400: policy is missing a condition`.

smecsia (approved):

Laboratory (approved):

PR #399, opened 2026-09-15
fix(security): deploy built images by digest on every cloud, not just Kubernetes

Cre-eD:
Every post-push security operation already runs against the immutable content digest — `resolveSecurityImageRef` rejects anything that is not `name@sha256:<64 hex>`, and scan, sign, verify, SBOM and provenance all use it. The runtime, however, was pointed at a tag.

So the artifact that is scanned, signed and verified was not the artifact the runtime is told to pull. Anyone able to push to the registry can move the tag after verification passes, and what runs is an image no check ever saw. Registry-level immutable tags help, but the guarantee only holds if the manifest reference itself carries the digest.

Three commits, smallest first.

### 1. Kubernetes: deploy by digest

`BuildAndPushImage` now returns `DeployImageRef` and the container spec uses it.

Strictness follows the stack's own posture. A stack with signing enabled fails closed when no digest is available — deploying a mutable reference would silently void the guarantee it asked for. A stack without signing keeps working on the tag and gets a warning, so consumers whose registry or provider does not report a digest are not broken.

### 2. Three corrections from review

- **Nothing is pushed during a preview**, so there is no digest to resolve. The strict path would have failed the plan rather than the deployment — the wrong moment and the wrong signal. The tag is returned quietly during a dry run; the check happens on the update, where it means something.
- **Strictness read `signing.enabled` alone**, but signing only runs when the top-level security flag is set too. A stack with security disabled and signing left on would have been treated as signed and failed closed for a guarantee it never had. It now reads the same pair the security operations do, with a test over all four combinations.
- **The fallback warning went to process stdout**, bypassing the Pulumi diagnostic stream the provisioning output is collected from — so the one case where the guarantee is knowingly given up was the case least likely to be seen. It goes through `ctx.Log.Warn` now.

### 3. AWS: ECS and Lambda were missed

The plumbing existed and was thrown away. `buildAndPushDockerImageV2` dropped `ImageOut.DeployImageRef`, so the ECS task definition took `docker.Image.ImageName` and the Lambda function took the same string as `ImageUri`. Both tags.

Same hole, in the two places nobody looked. On Lambda it is worse than on ECS: the function stores the resolved image, so a moved tag is picked up on the next update with no deployment at all.

`ECRImage.ImageName` stays the tag — it is what the stack exports and what a human reads. The new `DeployImageRef` field is what the container definition references.

## Why the AWS test is structural

The regression is structural, not logical: a new consumer wires up the field that is easy to reach instead of the one that is correct. A behavioural test would assert that a struct assignment happened.

So the test walks the package AST for the fields that decide what actually runs — `ecs.TaskDefinitionContainerDefinitionArgs.Image` and `lambda.FunctionArgs.ImageUri` — and requires each to be fed a `DeployImageRef`. The mirrored helpers image is on an explicit allowlist with its reason: it is built outside `BuildAndPushImage` and has no digest to use. A new runtime image reference fails the test until someone decides which case it is.

Verified by reintroducing the Lambda bug: the test names the file and the line. It also fails if the field names ever go stale, rather than passing by checking nothing.

## Blast radius

The deployed reference changes from `repo:tag` to `repo@sha256:...` on the next update of every stack that builds an image. Stack outputs are unchanged — the exported image name is still the tag.

A stack with signing enabled whose registry does not report a digest will now fail instead of deploying. That is the intended behaviour and it is the only breaking case; without signing, the tag path still works and warns.

### DIFF
--- a/pkg/clouds/pulumi/docker/build_and_push.go
+++ b/pkg/clouds/pulumi/docker/build_and_push.go
@@ -50,6 +50,11 @@ type Image struct {
 type ImageOut struct {
 	Image   *docker.Image
 	AddOpts []sdk.ResourceOption
+	// DeployImageRef is the reference a runtime should be pointed at. It is the
+	// immutable digest whenever one is available, so that what runs is what was
+	// scanned, signed and verified rather than whatever the tag points at by the
+	// time the runtime pulls.
+	DeployImageRef sdk.StringOutput
 }
 
 // BuildAndPushImage builds a Docker image, pushes it, and runs security
@@ -111,7 +116,11 @@ func BuildAndPushImage(ctx *sdk.Context, stack api.Stack, params pApi.ProvisionP
 		addOpts = append(addOpts, sdk.DependsOn([]sdk.Resource{res}))
 	}
 
-	return &ImageOut{Image: res, AddOpts: addOpts}, nil
+	return &ImageOut{
+		Image:          res,
+		AddOpts:        addOpts,
+		DeployImageRef: resolveDeployImageRef(res.RepoDigest, res.ImageName, signingEnabled(stack.Client.Security)),
+	}, nil
 }
 
 // executeSecurityOperations creates Pulumi resources for post-push security ops.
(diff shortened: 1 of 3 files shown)


smecsia (approved):

universe-ops (approved):

Cre-eD:
Superseded by #401, which carries this together with the commit-tag reuse that was stacked on top of it. The Kubernetes half of this landed separately in #395, so what remained here was the AWS part alone; #401 is that plus the reuse work, rebased on current main as two commits.

PR #400, opened 2026-09-15
feat(docker): reuse the image already under a commit tag instead of pushing over it

Cre-eD:
Stacked on #399 — the base retargets to `main` when that merges. Review only the last commit.

## The problem

Container builds are not reproducible. Rebuilding one commit tomorrow produces a different digest than it did today: timestamps move, base images drift, layer ordering changes.

The tag this toolchain generates is `YYYY.MM.DD-<sha7>`, so it names exactly one commit. Yet rerunning a deploy of that commit replaced the artifact under the tag with a different one. **Whatever was scanned, signed and verified on the first run is not what ships on the second.**

The same thing breaks registries with immutable tags outright: a second push under an existing tag is refused, so a rerun that changed nothing fails the deploy.

## What this does

With `imageBuild.reuseExistingCommitTag: true`, a deploy resolves the tag in the registry first. If it is there, the push is skipped and the digest already in the registry is what the security pipeline runs against and what the runtime is pointed at. A rerun of one commit deploys the identical artifact.

```yaml
schemaVersion: "1.0"

imageBuild:
  reuseExistingCommitTag: true   # default: false
```

## Which versions qualify

Only `YYYY.MM.DD-<7 hex>`. That shape names one commit; nothing else does.

| Version | Reused | Why |
|---|---|---|
| `2026.09.14-4fc2fda` | yes | names one commit |
| `latest` | no | moves |
| `2026.09.14-nogit` / `-nohash` / `-gitfail` | no | the commit could not be read |
| `v1.2.3`, branch names, anything from `VERSION` or `--deploy-version` | no | not commit-derived |

An empty version already becomes `latest` on every cloud, and both `VERSION` and `--deploy-version` accept arbitrary input, so reuse under them would serve a stale artifact for changed source.

## Failure handling

Only a **missing tag** means "build". Authentication failures, an unreachable daemon, transport errors and 5xx responses are returned as errors.

This is the part worth reviewing closely: a registry that cannot be read must not look like a registry with nothing in it. Conflating the two would push over an existing tag precisely when the check failed — the one case where the guarantee matters most. A digest returned in an unexpected shape is an error too, not a quiet fallback to the tag.

Tests cover each of those shapes separately, plus a registry host with a port (a colon that is not a tag separator) and the encoded-auth blob actually carrying the credentials.

## Why the resource graph does not change

The obvious implementation — skip `docker.NewImage` when the tag exists — makes the declared resource graph depend on registry state, so the next deploy that does build would delete and recreate.

Instead the image resource is still declared unconditionally and `SkipPush` carries the decision. Registry state never adds or removes a resource, and no aliases or resource-type migration are involved. The push was already skipped during a preview; that is now the same expression.

The build itself still runs. Skipping it too would mean moving build orchestration out of `docker.Image` and into a shell command, which loses Pulumi's diffing and puts a full build log into command state. The push is where the damage was.

`executeSecurityOperations` takes the digest as an argument now rather than reading it off the image resource, because on a reuse that field describes a local build that was never pushed.

## Blast radius

Off by default, and a no-op for any stack whose version is not commit-pinned. A stack that opts in and whose registry is unreadable fails the deploy instead of silently rebuilding — intended, and the only new failure mode.


smecsia (approved):

universe-ops (approved):

Cre-eD:
Superseded by #401. This was stacked on the branch of #399, which made it unreadable on its own; both are now one pull request against main.

